PR #64, opened 2017-10-02
updates to hydrograph refresh when options change

troddick:


jknewson (approved):
added MIT licence, should only merge to master when ready for production.

Issue #89, opened 2019-09-11
Add Manage Citations button near edit scenario button


kjacobsen16:
with https://github.com/USGS-WiM/NSSServices/issues/97

>Maybe the easiest/ straight forward way of handing this is adding another button next to add/edit scenario,  something like manage citations, which shows a modal of all citations (filtered when logged in etc) that shows the citation and which regression region they are associated to (I don't think that endpoint exists yet, but should be easy to set up). The delete button can be disabled if there is an association to any type of reg region. That way the user can easily see that they need to go to that reg-region and remove the citation before they can delete it

kjacobsen16:
@harper-wavra @amedenblik this doesn't have a service endpoint yet to see which regression region the citations are associated to, but setting up the button and modal, and maybe just pulling all citations for now is a good place to start

amedenblik:
I'm going to try to tackle this issue.

kjacobsen16:
Manage citations modal added - will need more work once services are set up

kjacobsen16:
@amedenblik the endpoint is now live! You can see/test it in the documentation here: https://test.streamstats.usgs.gov/docs/nssservicesv2/#/RegressionRegions/GET/RegressionRegions.

You call it exactly the same as the regular regression regions calls, but with the "bycitations" parameter sent through as true.

I don't think you've set up a service call yet for this app. You can start by looking at the regressionregions call when a user selects a region. Let me know if you need help, and I'll point you in the right direction :)

kjacobsen16:
Making a note here because I almost forgot.. This is waiting for the Angular update (#100) since the http requests will be different

amedenblik:
Removing myself from this issue since I am running low on hours. Harper and KJ, please let me know if you have questions.

kjacobsen16:
@harper-wavra I added the regression regions to the get all citations endpoint, so you should be able to loop through those now and add them to the modal. I'll take a look at your latest commits soon.

PR #104, opened 2020-01-09
Add Manage Citations button near Edit Scenario button

amedenblik:
Fixes issue #89. 

I set up a "Manage Citations" button and modal that shows all citations based on the filtering in the sidebar. Service endpoint does not exist yet, so the citations cannot actually be "managed" yet.


harper-wavra (approved):
Looks good to me!

src/app/shared/components/managecitations/managecitations.component.ts
```diff
@@ -0,0 +1,112 @@
...
+            }
+        });
+        this.modalSubscript = this._nssService.showManageCitationsModal.subscribe((show: boolean) => {
+            if (show) { this.showModal(); }
+        });
+        this._nssService.selectedRegion.subscribe(region => {
+            this.selectedRegion = region;
+            if (region && region.id) {this.getRegRegions(); }
+        });
+        this._nssService.getVersion.subscribe((v: string) => {
```
kjacobsen16: I think this can be taken out as well.

amedenblik: Which lines?

kjacobsen16 (changes requested):
This looks great! I left one comment - normally I wouldn't mention it, but this can be a learning experience with pull requests that you have to make changes on!

src/app/shared/components/managecitations/managecitations.component.ts
kjacobsen16: Oh yeah, that looks confusing. Starting at 59, that function where it subscribes to the appVersion. I don't think the appVersion is used. You can take that out of the variables as well.

kjacobsen16 (approved):
Looks good! 👍

PR #107, opened 2020-01-17
Fixes #105

### LINKED ISSUE #105: Save Add Scenario edits on close and add "Clear All" button to modal
Thinking back to yesterday's call, I think it would be helpful to save the fields people put into the Add Scenario modal somehow, so if they need to close out of it to check something their edits will be there when they reopen it until either they submit the scenario or clear it.

harper-wavra:


src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -171,11 +170,29 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
 
     showMathjax() {
         const exp = this.newScenForm.get('regressionRegions.regressions.equation').value;
+        if(exp == null){
+            this.equation = " ";
+            const equ = document.getElementById('mathjaxEq');
+            equ.style.visibility = 'hidden';
+            if (equ.firstChild) {equ.removeChild(equ.firstChild); }
+            equ.insertAdjacentHTML('afterbegin', '<span [MathJax]="equation">' + this.equation + '</span');
+        }else{
```
kjacobsen16: Can you walk me through the changes you made to this section? It looks like there's some repetition with the `else` portion below, and I'm wondering if that can be avoided.

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -171,11 +170,29 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
+            value: new FormControl(null, Validators.required)
+        }));
+    }  
+
+    fillError(id,index){
+        const errorList = <FormArray>this.newScenForm.get('regressionRegions.regressions.errors');
+        const error = errorList.at(index).value;
+        const selectedError = this.errors.find(x => x.id == id);
+        error.name = selectedError.name;
+        error.code = selectedError.code;
```
kjacobsen16: So this is interesting: somewhere along the way the code and name are getting lost. I'm guessing it's because you didn't put code and name in the error form group (in the addError() function), image is of the error object when adding a scenario
image

What this means is that we only need the id and value of the error, not the name and code, so we can remove that whole fillError section here, as well as the call in the .html, which I'll also comment on

src/app/shared/components/addscenario/addscenario.component.html
```diff
@@ -175,24 +178,23 @@ <h4 class="modal-title pull-left">New Scenario </h4>
...
-                            <div [id]="'error' + (errIndex + 1)" class="scenFormGroup">
-                                <!-- Error ID-->
+                            <div [formGroupName]=errIndex [id]="'error' + (errIndex + 1)" class="scenFormGroup">
+                                <!--Error ID-->
                                 <div class="form-group required">
-                                    <label class="req" [for]="errIndex">Error:</label>
-                                    <select class="form-control" [formControlName]="errIndex">
-                                        <option *ngFor="let e of errors" [ngValue]="e">{{e.name}}</option>
+                                    <label class="req" for="id">Error:</label>
+                                    <select class="form-control" formControlName="id" (ngModelChange)="fillError($event, errIndex)" (ngModel)="errors">
```
kjacobsen16: See comment in .ts file, the `(ngModelChange)="fillError(...)"` part can be removed

kjacobsen16 (changes requested):
Overall, looks good. I made some comments just about cleaning up some things (and one was partially my fault). Let me know if you have questions about my comments.

src/app/shared/components/addscenario/addscenario.component.ts
harper-wavra: image

I was getting a weird null character when I cleared the form so I was trying to fix that. Your right though, there was a lot of repetition. I tried to rewrite it better in the last commit.

kjacobsen16: Ah, ok. Thanks for explaining and for cleaning that up!

kjacobsen16 (approved):
Looks good, and from my tests seems to be working really well!

PR #109, opened 2020-02-06
HW 91

harper-wavra:
Adds clone scenario button
Removes parameter comment section


src/app/app.component.ts
```diff
@@ -51,6 +56,11 @@ export class AppComponent implements OnInit, OnDestroy {
     }
 
     ngOnInit() {
+        this._nssService.currentItem.subscribe(item => this.item = item);
+        this._nssService.currentRegRegion.subscribe(regRegion => this.regRegion = regRegion);
+        this._nssService.currentStatisticGroup.subscribe(statisticGroup => this.statisticGroup = statisticGroup);
+        this._nssService.currentClone.subscribe(clone => this.clone = clone);
+
```
kjacobsen16: I don't think these are needed since these items aren't used elsewhere in the app component.

src/app/mainview/mainview.component.ts
```diff
@@ -174,6 +180,13 @@ export class MainviewComponent implements OnInit, OnDestroy {
             if (region) { this.showRegion = true; }
             window.scrollTo(0, 0);
         });
+
+
+        this._nssService.currentItem.subscribe(item => this.item = item);
+        this._nssService.currentStatisticGroup.subscribe(statisticGroup => this.statisticGroup = statisticGroup);
+        this._nssService.currentRegRegion.subscribe(regRegion => this.regRegion = regRegion);
+        this._nssService.currentClone.subscribe(clone => this.clone = clone);
```
kjacobsen16: Same as above, I don't think these variables are actually used anywhere else in the mainview component.

src/app/mainview/mainview.component.ts
```diff
@@ -1205,8 +1218,14 @@ export class MainviewComponent implements OnInit, OnDestroy {
         popupWin.document.close();
     }
 
-    /////////////////////// Add/Edit/Delete Scenarios Section ///////////////////////////
+    /////////////////////// Add/Clone/Edit/Delete Scenarios Section ///////////////////////////
```
kjacobsen16: I wonder if we should break these apart for organizational purposes

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -152,6 +162,82 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
+        });
+
+        this.newScenForm.patchValue({
+            statisticGroupID: this.statisticGroup.toString(),
+            regressionRegions: {
+                ID: this.regRegion.id.toString(),
+                regressions:{
+                    ID: this.itemParameters.id.toString(),
+                    equation: this.itemParameters.equation.toString(),
+                    equivalentYears: this.itemParameters.equivalentYears.toString(),
```
kjacobsen16: I had issues with this when cloning one of the regressions from the test region. I think this is because I didn't get it a value for equivalent years when I created it. Instead of it being 0, it's coming back as undefined so it can't be converted using `.toString()` here, causing a lot of the cloning to fail.

I also noticed that when I clone regressions that do have values for equivalent years it gets reset to 0 when cloned. I didn't dive into it too much, but I noticed this occurred even if I remove the `.toString()` method. It might be getting overwritten at some point.

src/app/app.component.ts
```diff
@@ -29,6 +30,10 @@ export class AppComponent implements OnInit, OnDestroy {
     public LoginForm: FormGroup;
     public modalRef;
     public title;
+    public item: any;
+    public regRegion: any;
+    public statisticGroup: any;
+    public clone: boolean;
```
kjacobsen16: I don't think these are needed since these items aren't used elsewhere in the app component.

src/app/mainview/mainview.component.ts
```diff
@@ -29,6 +29,7 @@ import { NgForm, FormGroup, FormBuilder, FormControl, Validators } from '@angula
 import { Predictioninterval } from 'app/shared/interfaces/predictioninterval';
 import { URLSearchParams } from '@angular/http';
 import { Citation } from 'app/shared/interfaces/citation';
+import { element } from '@angular/core/src/render3';
```
kjacobsen16: Looks like this isn't used anywhere

src/app/mainview/mainview.component.ts
```diff
@@ -48,6 +49,10 @@ export class MainviewComponent implements OnInit, OnDestroy {
     @ViewChild('CitationForm') citationForm;
     public newCitForm: FormGroup;
     public title: string;
+    public statisticGroup: any;
+    public item: any;
+    public regRegion: any;
+    public clone: boolean;
```
kjacobsen16: Same as above, I don't think these variables are actually used anywhere else in the mainview component.

src/app/mainview/mainview.component.ts
```diff
@@ -1361,6 +1380,29 @@ export class MainviewComponent implements OnInit, OnDestroy {
...
+        this._nssService.changeRegRegion(regr);
+    }
+
+    newStatisticGroup(sg){
+        this._nssService.changeStatisticGroup(sg);
+    }
+
+    newClone(x){
+        this._nssService.changeClone(x);
+    }
```
kjacobsen16: These probably don't need their own function, since they're one line of code.

src/app/mainview/mainview.component.ts
```diff
@@ -1361,6 +1380,29 @@ export class MainviewComponent implements OnInit, OnDestroy {
...
+
+    public cloneRowClicked(sg, i, regr) {
+        this.newItem(i);
+        this.newRegRegion(regr);
+        this.newStatisticGroup(sg);
+        this.showCloneScenarioModal();
```
kjacobsen16: Could we possibly combine these into one item to send through to the add scenario component, since we combine them to get the newScenForm anyway? Then we'd only need one subscription in the addscenario component. So, if the user chooses to clone the scenario, we'd combine it into one object and send that to the addScenario modal (using the service and a subject/subscription), which would tell it that it should use that clone. That can then trigger the modal to open, getting rid of lines 1227-1229.

src/app/shared/components/addscenario/addscenario.component.html
```diff
@@ -52,11 +52,10 @@ <h4 class="modal-title pull-left">New Scenario </h4>
                         <div class="input-invalid-warning" *ngIf="!newScenForm.get('regressionRegions.regressions.equation').valid && newScenForm.get('regressionRegions.regressions.equation').dirty">Expression
                             is required</div>
                     </div>
-
                     <!-- Equivalent Years -->
                     <div class="form-group">
                         <label class="req" for="equivalentYears">Equivalent Years (optional):</label>
-                        <input class="form-control" (keypress)="_keyPress($event)" type="number" formControlName="equivalentYears" >
+                        <input class="form-control" (keypress)="_keyPress($event)" type="number" [value]="0" formControlName="equivalentYears" >
```
kjacobsen16: I wonder if this edit is causing the issues I mentioned in a different spot where the existing equivalentYears aren't showing up when cloning a scenario. maybe we want to check if the value exists and set it to 0 if not.

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -6,7 +6,7 @@
 // authors:  Tonia Roddick USGS Wisconsin Internet Mapping
 // purpose: modal used to show about information
 
-import { Component, ViewChild, OnInit, OnDestroy } from '@angular/core';
+import { Component, ViewChild, OnInit, OnDestroy, Input } from '@angular/core';
```
kjacobsen16: This isn't being used, probably just forgot to remove it :)

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -81,6 +85,11 @@ export class AddScenarioModal implements OnInit, OnDestroy {
     }
 
     ngOnInit() {
+        this._nssService.currentItem.subscribe(item => this.itemParameters = item);
+        this._nssService.currentRegRegion.subscribe(regRegion => this.regRegion = regRegion);
+        this._nssService.currentStatisticGroup.subscribe(statisticGroup => this.statisticGroup = statisticGroup);
+        this._nssService.currentClone.subscribe(clone => this.clone = clone);
+
```
kjacobsen16: As I mentioned in the mainview.component, if we combine everything into one object (say item), we'll only need one subscription here.

src/app/shared/interfaces/parameter.ts
```diff
@@ -9,6 +9,7 @@ export interface Parameter {
     unitType: Unittype;
     value: number;
     limits: Limit;
+    comments: string;
```
kjacobsen16: ha, we decided we don't need this anymore so it can be removed, my bad

src/app/shared/interfaces/predictioninterval.ts
```diff
@@ -5,4 +5,6 @@ export interface Predictioninterval {
     xiRowVector?: string;
     XIRowVector?: string;
     covarianceMatrix: string;
+    lowerBound: number;
+    supperBound: number;
```
kjacobsen16: minor typo

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -152,6 +162,82 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
             }
         );
+
+        if (this.clone == true){
+            this.clearScenario();
+            this.cloneSenario();
+        }  
+    }
+
+    cloneSenario(){
```
kjacobsen16: minor typo here, "Scenario", not "Senario", make sure it's updated anywhere it calls it (e.g. line 168). Using that Code Spell Checker extension in VScode might catch that.

src/app/shared/services/app.service.ts
```diff
@@ -61,10 +61,48 @@ export class NSSService {
...
+    }
+
     // show the add scenario modal in the mainview
     public get showAddScenarioModal(): any {
         return this._showHideAddScenarioModal.asObservable();
     }
+    // show the add scenario modal in the mainview
+    public get showCloneScenarioModal(): any {
+        return this._showHideCloneScenarioModal.asObservable();
+    }
```
kjacobsen16: A lot of this will be cleaned up if we decide to just send one object from the mainview when we clone

kjacobsen16 (changes requested):
I made a lot of comments, but this was a big task, and there's a lot that went into it. Angular is a big learning curve, so it takes some getting used to to figure out what is the most efficient way to do things.

I made a lot of inline comments, hopefully those are easy to follow. Something to note is that if you use VSCode, you can double-check your changes in the source control (let me know if you don't know what that is). I mainly just glance at them to make sure I didn't forget to remove something I didn't end up using.

I tried cloning a regression in the test region by _just_ changing the regression type and adding parameter and expected values, but it's not updating the regression type for some reason. This should definitely be fixed. I would go through and make sure all changes you make are being sent through when you hit the "Create" button. diffchecker.com can be used to compare the scenario you're cloning to your cloned and edited scenario.

I like the feel of it and it's such a nice feature to have! Let me know if you need help finding the solutions to these issues.

src/app/mainview/mainview.component.ts
harper-wavra: Do you want me to try and break these up?

kjacobsen16: Sure - if it makes sense to do so. It would just be adding a comment noting each section and then grouping any functions related to it together. That was my thought without looking at how many functions there are, just seemed like a lot to be grouped together.

src/app/mainview/mainview.component.ts
harper-wavra: Do you know of an easy way to combine these without overriding things like ID, name, code ,ect?

kjacobsen16: I'm not sure what you mean by that, I'd have to look into the code better. What I would start with is looking at where you're combining the scenario in your clonescenario function in the addscenario component and trying to replicate that.

Or, you could just create one object and sending each of these items through as a property, e.g. {newItem: i, newRegRegion: regr, newStatisticGroup: sg}. This is probably not the best "Angular" way to do it, but just showing you a way to combine it and remove all these calls/subscriptions.

harper-wavra:
I add those changes. Let me know what you think. I tested cloning a regression in the test region in South Dakota by just changing the regression type and adding parameter expected value and expected value, but my regression type did update. Could you send me what you used to get the bug?

image

src/app/mainview/mainview.component.ts
```diff
@@ -29,7 +29,8 @@ import { NgForm, FormGroup, FormBuilder, FormControl, Validators } from '@angula
 import { Predictioninterval } from 'app/shared/interfaces/predictioninterval';
 import { URLSearchParams } from '@angular/http';
 import { Citation } from 'app/shared/interfaces/citation';
-import { element } from '@angular/core/src/render3';
+import { connectableObservableDescriptor } from 'rxjs/internal/observable/ConnectableObservable';
```
kjacobsen16, line 32: This looks like an accidental import. For now this is fine, but it's good to keep an eye out for lines that show up as transparent in VSCode, since that means they're not used.

src/app/mainview/mainview.component.ts
```diff
@@ -1218,17 +1209,22 @@ export class MainviewComponent implements OnInit, OnDestroy {
...
+    public showCloneScenarioModal() {
         this._nssService.setAddScenarioModal(true);
     }
 
-    public showCloneScenarioModal() {
-        this.newClone(true);
-        this._nssService.setAddScenarioModal(true);
+    newCloneScenario(cloneScen){
+        this._nssService.changeItem(cloneScen);
     }
```
kjacobsen16: If this is only used once and is only one line, it doesn't need to be a separate function. You can just put line 1218 in place of line 1223 and remove this function.

src/app/mainview/mainview.component.ts
```diff
@@ -1218,17 +1209,22 @@ export class MainviewComponent implements OnInit, OnDestroy {
...
 
+    public cloneRowClicked(statisticGroupID, r, rr) {
+        this.cloneScen={r,rr,statisticGroupID};
```
kjacobsen16, line 1222: I think this is fine for now, but this can be optimized a little, and could even be a one line function. If `this.cloneScen` doesn't get used elsewhere, you can just say `this._nssService.changeItem({r, rr, statisticGroupID })`. 

Even better, you could combine them into an object in the html. You don't need to send `statisticgroupID, r, rr` through when you're just going to combine them into an object to send through right away. Instead, in the html where you call this function you can say `(click)="cloneRowClicked({r, rr, s.statisticGroupID})"`. 

Also, instead of triggering `showCloneScenarioModal`, since you're only showing it when you hit clone, you can just call `this.showModal()` inside your subscriber in the addscenario modal, line 89.

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -16,6 +16,7 @@ import { ConfigService } from 'app/config.service';
 import { Scenario } from 'app/shared/interfaces/scenario';
 import { ToasterService } from 'angular2-toaster';
 import { AuthService } from 'app/shared/services/auth.service';
+import { SSL_OP_SSLEAY_080_CLIENT_DH_BUG } from 'constants';
```
kjacobsen16, line 19: Probably an accidental import

kjacobsen16 (approved):
Looks good! I left some comments on things that could be optimized, so if you want to fix those you can push those. Otherwise, it's working well so I'll go ahead and merge.

PR #121, opened 2020-02-21
Fixes #116

### LINKED ISSUE #116: Hide Area-Averaged region
While we figure out some changes to area averaging, Pete M. wants us to hide the "Area-Averaged" region from the results.

Looks like this was built in at some point - https://github.com/USGS-WiM/NSS/blob/dev/src/app/mainview/mainview.component.html#L138, if you uncomment the `ngIf` statement, that should do it.

amedenblik:
Fixes #116 

Remove area-averaged region results (for now)


kjacobsen16 (changes requested):
Looks good, but I just realized we can also take out the % Area weight prompts for now (just comment out, we'll need them again later)

harper-wavra (approved):

amedenblik (commented):
Made requested changes

kjacobsen16 (approved):

PR #149, opened 2020-03-23
fixes #139

### LINKED ISSUE #139: When adding new lookups close module
When you add a new error, unit type, ect in the settings. Close the module when the user clicks create.

harper-wavra:
Setting modals now close when user clicks "create"


kjacobsen16 (approved):
Looks good, this is much better!

PR #156, opened 2020-04-01
Fixes #145

### LINKED ISSUE #145: Incorporate Freshdesk For Reporting Bugs

harper-wavra:
In about module added help tab to send in a FreshDesk ticket.

### DIFF
--- a/src/app/shared/components/about/about.component.ts
+++ b/src/app/shared/components/about/about.component.ts
@@ -51,7 +51,7 @@ export class AboutModal implements OnInit, OnDestroy {
         ) {
         this.form = this.fb.group({
             name:"",
-            avatar: [null]
+            file: [null]
         })
         this.freshDeskTicket = new freshDeskTicket();
     }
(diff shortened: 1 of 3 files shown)


src/app/shared/components/about/about.component.ts
```diff
@@ -17,22 +25,47 @@ import { NSSService } from 'app/shared/services/app.service';
...
+        private _modalService: NgbModal, 
+        private _toasterService: ToasterService,
+        ) {
+        this.newTicketForm = fb.group({
+            'email': new FormControl(null, Validators.required),
+            'subject': new FormControl(null, Validators.required),
+            'description': new FormControl(null, Validators.required),
+            'attachment': new FormControl(null)
+        });
+        this.freshDeskTicket = new freshDeskTicket();
```
kjacobsen16: I think having the form and the interface is unnecessary. I think the `freshDeskTicket` object is just used for the file. Can we merge that with the `newTicketForm`, and access it from there, like the other properties in the `newTicketForm`? Then we can remove that interface file (and imports) too. You should be able to set the value in the uploadFile() function in a similar way that we're clearing the value in the removeFile() function

kjacobsen16 (changes requested):
Working great! I left a comment about some code improvements

kjacobsen16 (approved):
Looks good

PR #158, opened 2020-04-08
Fixes #129

### LINKED ISSUE #129: Add loader to main page
We should add a loader on the main page until the regions list is returned (and possibly other places). Sometimes I get to the app and the dropdown doesn't work because the region list hasn't loaded yet.

An example: https://test.wim.usgs.gov/liqwids/?characteristic=Nitrate,Nitrogen&minResults=1, we might be able to replicate this, or have Mitch help out.

harper-wavra:
I added the loader from LIQWIDS to a couple places within the app. I added it to the main page. It waits for the region list to be returned. I also added it when you chose a new region.  It shows up when you log in and log out. Finally it shows when you use the sidebar buttons ("Compute" and "Reset Region Filters".)


kjacobsen16 (approved):
Looks great - a big improvement, especially with the regions list!

PR #160, opened 2020-04-13
Fixes #132

### LINKED ISSUE #132: Enable assigning new Region on Clone Scenario
Add option to change the region for a scenario when cloning, which means the regression regions will also need to update

harper-wavra:
Now able to clone an equation for a different region.

### DIFF
--- a/src/app/shared/components/addscenario/addscenario.component.html
+++ b/src/app/shared/components/addscenario/addscenario.component.html
@@ -13,8 +13,8 @@
 
             <!-- Clone - choose region -->
             <div class="form-group" *ngIf="clone==true">
-                <label class="req" for="regionid">Region:</label>
-                <select class="form-control" formControlName="regionid">
+                <label class="req" for="region">Region:</label>
+                <select class="form-control" formControlName="region">
                     <option *ngFor="let r of regions" [ngValue]="r">{{r.name}}</option>
                 </select>
             </div> 
--- a/src/app/shared/components/addscenario/addscenario.component.html
+++ b/src/app/shared/components/addscenario/addscenario.component.html
@@ -1,7 +1,8 @@
 <ng-template #addScenario let-c="close" let-d="dismiss">
     <div class="modal-header">
         <div class="col-xs-12">
-            <h4 class="modal-title pull-left">New Scenario </h4>
+            <h4 class="modal-title pull-left" *ngIf="clone==true"> Clone Scenario </h4>
+            <h4 class="modal-title pull-left" *ngIf="clone==false"> New Scenario </h4>
             <!--close and save-->
             <button class="close pull-right" (click)="c('closed')">x</button><br><br>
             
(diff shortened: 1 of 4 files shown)


src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -42,8 +43,9 @@ export class AddScenarioModal implements OnInit, OnDestroy {
     public cloneParameters: any;
     public regRegion: any;
     public statisticGroup: any;
-    public clone: boolean;
+    public clone: boolean = false;
```
kjacobsen16: This is fine, but for future reference you don't need to type set it if you're setting an initial value. So just saying `public clone = false;` would have been fine :)

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -42,8 +43,9 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
     public selectedRegion;
+    public orgRegion;
```
kjacobsen16: Is this short for "original region"? maybe this should be changed to `originalRegion`

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -53,6 +55,7 @@ export class AddScenarioModal implements OnInit, OnDestroy {
         private _settingsService: SettingsService, private _configService: ConfigService, private _toasterService: ToasterService,
         private _authService: AuthService) {
         this.newScenForm = _fb.group({
+            'regionid': new FormControl(null, Validators.required),
```
kjacobsen16: You're actually setting this to the whole region, so the property should probably be named `region` instead

src/app/shared/components/addscenario/addscenario.component.html
```diff
@@ -1,21 +1,31 @@
...
     <div class="modal-body">
         <form [formGroup]="newScenForm">
+
+            <!-- Clone - choose region -->
+            <div class="form-group" *ngIf="clone==true">
+                <label class="req" for="regionid">Region:</label>
+                <select class="form-control" formControlName="regionid">
+                    <option *ngFor="let r of regions" [ngValue]="r">{{r.name}}</option>
+                </select>
+            </div>
```
kjacobsen16: I think is fine, but I'm wondering if it could have been done outside of the form (so placed after line 11 instead), and just hooked up to adjust the `selectedRegion` variable, so you don't have to add it to, and remove it from, the form.

kjacobsen16 (changes requested):
Works well. I left a couple comments. There is one issue where the region sort of updates after you clone to a new region. The mainview component shows the new region, but the sidebar doesn't show it. It seems like either it's not sending the new region through correctly, or the sidebar isn't subscribing to the `selectedRegion` observable from the app.service correctly.

harper-wavra:
In response to "The mainview component shows the new region, but the sidebar doesn't show it", after cloning to a new region do we want the mainview and sidebar to show the new region or the original region?

kjacobsen16:
We should show the new region so they can see the new equation. I checked this with Ryan and he agrees.

kjacobsen16 (approved):
Works great - I think this will be an exciting one to have finished!

PR #168, opened 2020-04-20
Fixes #112

### LINKED ISSUE #112: Filters not sticking on add scenario
If you have a filter set then add a scenario, it'll reload the scenarios without the filters

harper-wavra:
Saves filters when adding scenario


kjacobsen16 (changes requested):
Looks like the filter is there, but it's not showing the new scenario when added. Like, if you filter by a statistic group, and add a new scenario within that statistic group, it will only show the old ones.

What might work better is saving the filters and re-querying when a new scenario is added.

I noticed the filters not sticking is also an issue when you add a regression region, or edit or delete a regression, so you should be able to replicate the fix through all those.

harper-wavra:
Creating new pull request for issue.

Issue #164, opened 2020-04-20
Add arrows to dropdown boxes


harper-wavra:
Add arrow in right corner of all drop down boxes in order to indicate it is a drop down box.

From Pete's Feedback 4/16/20

PR #170, opened 2020-04-22
Added dropdown arrows #164

NathanKrawza:
After hours of tinkering... It was just a simple override in the main css file to add the dropdown arrow for every dropdown (without the arrow) in the application.


kjacobsen16 (approved):
Ha, love it when that happens. Looks great!

PR #172, opened 2020-04-24
fixes #148

### LINKED ISSUE #148: Issue with Error handling of Add Scenario function
I think this is an issue with how the error object gets sent to the `outputWimMessages()` function after we updated to the latest version of angular. We should make sure we check all places the function gets called.

NathanKrawza:
Fixed the errors so that they actually pop up


kjacobsen16 (approved):
Appears to be working well 👍

PR #173, opened 2020-04-24
Fixes #112

### LINKED ISSUE #112: Filters not sticking on add scenario

harper-wavra:
Saves filters and requerys when needed. I added fix to a few different spots.

### DIFF
--- a/src/app/mainview/mainview.component.ts
+++ b/src/app/mainview/mainview.component.ts
@@ -53,9 +53,12 @@ export class MainviewComponent implements OnInit, OnDestroy {
     public selectedRegion;
     private navigationSubscription;
     public selectedRegressionRegion: Array<Regressionregion>;
+    public tempRegressionRegion: Array<Regressionregion>;
+    public tempselectedStatisticGrp: Array<Statisticgroup>;
     public get selectedStatisticGrp(): Array<Statisticgroup> {
         return this._nssService.selectedStatGroups;
     }
+    public tempselectedRegType: Array<Regressiontype>;
     public get selectedRegType(): Array<Regressiontype> {
         return this._nssService.selectedRegressionTypes;
     }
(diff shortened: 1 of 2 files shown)


src/app/mainview/mainview.component.ts
```diff
@@ -53,9 +53,12 @@ export class MainviewComponent implements OnInit, OnDestroy {
     public selectedRegion;
     private navigationSubscription;
     public selectedRegressionRegion: Array<Regressionregion>;
+    public tempRegressionRegion: Array<Regressionregion>;
```
kjacobsen16: This is a minor thing, but I recommend keeping similar variables consistent, and capitalizing each word (camelCase), so in that case you would name it `tempSelectedRegressionRegion` and the next line `tempSelectedStatisticGroup`

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -384,6 +403,9 @@ export class AddScenarioModal implements OnInit, OnDestroy {
         this._settingsService.postEntity(scen, this.configSettings.scenariosURL + '?statisticgroupIDorCode=' + scen.statisticGroupID)
             .subscribe((response: any) => {
                 this._nssService.setSelectedRegion(this.selectedRegion);
```
kjacobsen16: I noticed that when you add/clone a scenario, for a second it shows the entire scenario, then hits the filters. Since the region is already selected, you should be able to remove this line and it should re-query properly with your `requeryFilters` function.

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -49,7 +52,17 @@ export class AddScenarioModal implements OnInit, OnDestroy {
     public addPredInt = false;
     public modalRef;
     public loggedInRole;
-
+    public selectedRegressionRegion: Array<Regressionregion>;
+    public tempRegressionRegion: Array<Regressionregion>;
+    public tempselectedStatisticGrp: Array<Statisticgroup>;
```
kjacobsen16: Same issues here with consistent naming

kjacobsen16 (changes requested):
I left a comment in the add scenario component - if you remove setting the selected region where you're re-querying, it'll avoid that issue.  This should be done in several places.

kjacobsen16 (approved):

PR #176, opened 2020-04-29
fixes #171

### LINKED ISSUE #171: Deleting wrong parameter/variable
When creating New Scenario, if you have more than one parameter/variable and click  the X to the left of variable it deletes variable 1 and moves all the variables up one spot.

For example, if I have these three variables and delete variable 3
image

It deletes variable 1 and moves variables 2 and 3 up to spots 1 and 2
image

NathanKrawza:
small change in the html fixed the issue


kjacobsen16 (approved):
Works great!

PR #177, opened 2020-04-29
Nk 174

NathanKrawza:
Changed image link to updated usgs logo. Had to add custom sizing/styling to the image to make it fit right. I'm not sure if this will have mobile implications? Let me know if there is a better way of doing this haha.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -4,8 +4,8 @@
 <nav class="wim-nav hidden-print">
     <div class="container" style="width: 100%;">
         <button class="mobile-nav-toggle" (click)="toggleSidebar()"></button>
-        <a href="https://usgs.gov" class="nav-branding">
-            <img src="https://wim.usgs.gov/website-assets/usgs-logo.png" />
+        <a href="https://usgs.gov">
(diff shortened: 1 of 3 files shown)


src/app/app.component.html
```diff
@@ -4,8 +4,8 @@
 <nav class="wim-nav hidden-print">
     <div class="container" style="width: 100%;">
         <button class="mobile-nav-toggle" (click)="toggleSidebar()"></button>
-        <a href="https://usgs.gov" class="nav-branding">
-            <img src="https://wim.usgs.gov/website-assets/usgs-logo.png" />
+        <a href="https://usgs.gov">
+            <img src="https://wim.usgs.gov/src/images/usgs.png" height="60px" width="150px" style="margin-top: 6px"/>
```
kjacobsen16: Generally, we try not to do inline styling (in the html).

Can you add a class or id and put this in a css file instead?

kjacobsen16 (changes requested):
One code change, otherwise it looks good!

kjacobsen16 (approved):
👍

PR #179, opened 2020-05-04
Fixes #102

### LINKED ISSUE #102: Incorporate feature to upload/preview hydrologic region polygons
Will discuss on call on 1/7, this should be added to the "Add Hydrologic Region" interface (was called Add Regression Region, not sure if it's updated).

We need the user to be able to upload a polygon of the hydrologic region, preview it to verify it, and submit. Eventually, this will send the polygon along to the services, but that endpoint does not exist yet. We'll also eventually want to allow the user to edit existing polygons.

Sample file to come.

harper-wavra:
I know this will probably have a lot of changes/clean up but I think it is working mostly correctly right now. I ran through adding/editing/cloning/deleting and those seemed to be working.


src/app/mainview/mainview.component.ts
```diff
@@ -600,7 +601,7 @@ export class MainviewComponent implements OnInit, OnDestroy {
             this.regTypes = res;
         });
         this.modalSubscript = this._nssService.showAddRegRegionModal.subscribe((show: boolean) => {
-            if (show) { this.showNewRegressionRegionForm(); }
+       //     if (show) { this.showNewRegressionRegionForm(); }
         });
```
kjacobsen16: I'm guessing this is commented out because we moved the add regression region form to its own component. If so, this section can be deleted.

src/app/mainview/mainview.component.ts
```diff
@@ -1279,19 +1278,26 @@ export class MainviewComponent implements OnInit, OnDestroy {
...
     }
 
-    public getCitations() {
-        this._settingsService.getEntities(this.configSettings.citationURL)
-            .subscribe(res => {
-                this.citations = res;
-            });
-    }
-
     public saveCitation(c) {
```
kjacobsen16: This function can probably be removed, since we are no longer editing citations in the main view, check other citation-related functions too to make sure they're needed. (this is for the `saveCitation()` function)

src/app/mainview/mainview.component.ts
```diff
@@ -1587,30 +1594,38 @@ export class MainviewComponent implements OnInit, OnDestroy {
...
+        // this.modalRef.result.then(
+        //     result => {
+        //         // this is the solution for the first modal losing scrollability
+        //         if (document.querySelector('body > .modal')) {
+        //             document.body.classList.add('modal-open');
+        //         }
+        //         if (result) {this.cancelCreateRegression(); }
+        //     },
+        //     reason => {if (reason) {this.cancelCreateRegression(); }}
+        // );
```
kjacobsen16: I think this can all be removed. This should all be in the reg region component now, since it's used for opening the modal.

src/app/mainview/mainview.component.ts
```diff
@@ -1587,30 +1594,38 @@ export class MainviewComponent implements OnInit, OnDestroy {
...
     }
 
     private cancelCreateRegression() {
```
kjacobsen16: I think the `cancelCreateRegression` and `createNewRegression` functions can be removed, since they were left over from the add regression region modal when it was in the mainview instead of its own component. There may be other functions as well.

src/app/mainview/mainview.component.ts
```diff
@@ -1667,6 +1688,11 @@ export class MainviewComponent implements OnInit, OnDestroy {
         );
     }
 
+    public uploadNewPolygon(rr) {
+        // Upload new polygon
+        // TODO: Send polygon to services
+    }
+
```
kjacobsen16: Can be deleted

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,422 @@
...
+  public CloseResult: any;
+  private modalSubscript;
+  public appVersion: string;
+  public regressionRegions;
+  public statisticGroups;
+  public regressionTypes;
+  public variables;
+  public unitTypes;
+  public equation;
+  public errors;
```
kjacobsen16: It looks like a lot of these were pulled from the add scenario modal. We don't need any of these in the add regression region modal.

There are more below as well, e.g. `addPredInt`. I'd just look over this carefully and check through the functions to make sure they're all needed, and used. A lot of these variables aren't used, even though they don't show up slightly transparent for me like they normally would.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,422 @@
...
+      'Satellite': tileLayer_satellite
+    };
+
+    // Add basemap control to the map
+    L.control.layers(baseMaps).addTo(this.map);
+
+  }
+
+  public getRegRegions() {
+    // moving to own function for when new regression region is added
```
kjacobsen16: This comment should be reworded, since we don't really know what it was moved from. I'm assuming this just means the regression regions get re-queried after a new one is added. We should make sure it isn't re-querying in another place (like resetting the selected region or some other way).

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,422 @@
...
+    }
+  }
+
+  public showAddRegRegion() {
+    const addRegRegForm: AddRegressionRegion = {
+      show: true,
+      regRegionID: null
+  }
+    this._nssService.setAddRegressionRegionModal(addRegRegForm);
+  }
```
kjacobsen16: This looks exactly like the call from in the mainview component. I don't know why it's in this component. I would assume we'd use it to invoke this component from another component. It might be another function that was carried over from the add scenario component by accident.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,422 @@
...
+    if (rr) { // edit existing regression region
+    this._loaderService.showFullPageLoad();
+    this.addRegReg = false;
+    this._settingsService.getEntities(this.configSettings.regRegionURL + '/' + rr + '?getpolygon=true')
+      .subscribe((res) => {
+        this.selectedRegRegion = res;
+        this.newRegRegForm.controls['name'].setValue(this.selectedRegRegion.name);
+        this.newRegRegForm.controls['description'].setValue(this.selectedRegRegion.description);
+        this.newRegRegForm.controls['code'].setValue(this.selectedRegRegion.code);
+        this.newRegRegForm.controls['location'].setValue(this.selectedRegRegion.location);
```
kjacobsen16: I wonder if this could be improved on. Something to think about.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,422 @@
...
+          .subscribe(res => {
+              this.selectedCitation = res;
+              this.newCitForm.controls['title'].setValue(this.selectedCitation.title);
+              this.newCitForm.controls['author'].setValue(this.selectedCitation.author);
+              this.newCitForm.controls['citationURL'].setValue(this.selectedCitation.citationURL);
+          });
+        } else {
+          this.addCitation = false;
+        }
+        if (this.selectedRegRegion.location) {
```
kjacobsen16: I think adding some comments or extra lines in this whole `showNewRegressionRegionForm` function would help improve readability, especially in places like this.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,422 @@
...
+  ngOnDestroy() {
+    this.modalSubscript.unsubscribe();
+  }
+
+
+  // show add regression region modal
+  public showNewRegressionRegionForm(rr?) {
+    // shows form for creating new regression or editing regression 
+    if (rr) { // edit existing regression region
+    this._loaderService.showFullPageLoad();
```
kjacobsen16: The lines within this if statement need to be indented.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,422 @@
...
+    this._loaderService.hideFullPageLoad();
+  }
+
+  public loadFile() {
+    this._loaderService.showFullPageLoad();
+    this.input = document.getElementById('fileinput');
+    if (!this.input.files[0]) {
+      alert("Please select a file.");
+    }
+    else {
```
kjacobsen16:
```suggestion
    } else {
```

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,422 @@
...
+                }, error => {
+                  if (this._settingsService.outputWimMessages(error)) { return; }
+                  this._toasterService.pop('error', 'Error creating Citation', error._body.message || error.statusText);
+                });
+            } else if (this.addCitation) {
+                this.createNewCitation(this.selectedRegRegion);
+            } else {
+                this._nssService.setSelectedRegion(this.selectedRegion);
+                this.cancelCreateRegression();
+            }
```
kjacobsen16: Again, adding some comments would improve readability.

src/app/shared/components/managecitations/managecitations.component.ts
```diff
@@ -62,6 +62,9 @@ export class ManageCitationsModal implements OnInit, OnDestroy {
             this.getRegRegions(); // get list of regression regions for the region
         });
         this.modalElement = this.manageCitationsModal;
+
+        // Subscribe to server with '?bycitation=true'
+        // Copy settingservice getEntities on regions.component.ts file
```
kjacobsen16, line 73: I think this is supposed to be for the manage citations branch (AM-89), just a note in case you start working on that branch again before this is merged into dev/back into the AM-89 branch.

src/app/shared/services/app.service.ts
```diff
@@ -92,8 +93,10 @@ export class NSSService {
         return this._showHideManageCitationsModal.asObservable();
     }
     // -+-+-+-+-+-+-+-+-+ add regression region modal -+-+-+-+-+-+-+-+
-    private _showHideAddRegressioRegionModal: Subject<boolean> = new Subject<boolean>();
-    public setAddRegressionRegionModal(val: any) {
+    private addregressionregion: AddRegressionRegion;
+    private _showHideAddRegressioRegionModal: Subject<AddRegressionRegion> = new Subject<AddRegressionRegion>();
+    public setAddRegressionRegionModal(val: AddRegressionRegion) {
+        this.addregressionregion = val;
```
kjacobsen16: I don't know why this is here, but I'm guessing this can be deleted, as well as line 66 where this variable is created.

src/app/shared/services/geojson.service.ts
```diff
@@ -0,0 +1,21 @@
...
+  readFileContent(file: File) {
+    let fileReader: FileReader = new FileReader();
+    fileReader.readAsArrayBuffer(file);
+    return Observable.create(observer => {
+      fileReader.onloadend = () => {
+        observer.next(fileReader.result);
+        observer.complete();
+      };
+    });
+  }
```
kjacobsen16: I wonder if this function can just be moved to a difference service file so we can delete this service file.

kjacobsen16 (changes requested):
As we expected, there's a lot of cleanup, a lot of it due to reorganization of the code. However, there are a couple big items:
- [x] We should add a loader when creating new regression region (it's slow if you're uploading a polygon)
- [x] Existing polygons aren't showing up when editing. To replicate, go to MN -> Edit scenarios -> Edit region A, should be the same as the test region zip I gave you, but is not showing up on the map. I'd start by getting the polygon for that region and running it through something like mapshaper.org to make sure there's nothing wrong with the geojson. Also by sticking a break point in the function that puts the geojson on the map. It's possible this is a services issue, but let's see what we can figure out client side first.
- [x]  Clicking "add a citation" doesn't work when editing scenarios

Thanks!

src/app/shared/services/geojson.service.ts
harper-wavra: Moved to loader.service, let me know if you think there is a better service to move it to.

kjacobsen16: The loader service should just be for the loader. The app service is probably the best one.

kjacobsen16:
@harper-wavra your merge from dev undid a lot of work in the mainview component. I'm guessing there were merge conflicts you accidentally took from dev instead.

kjacobsen16:
@harper-wavra alright I fixed that merge, and I'm pushing soon to the services so you'll have the projection working for all the polygons. One change: `getPolygon` was changed to `includeGeometry`. I believe you only call that once.

I also noticed the mainview component has an old function `showNewRegressionRegionForm` that's not being called anywhere. It can be removed.

src/app/mainview/mainview.component.ts
```diff
@@ -1660,7 +1578,7 @@ export class MainviewComponent implements OnInit, OnDestroy {
                 } else {
                     this._settingsService.outputWimMessages(res); 
                 }
-                this.cancelCreateRegression();
+                //this.cancelCreateRegression();
```
kjacobsen16: I think this can be removed instead of commented out

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,437 @@
...
+              private _configService: ConfigService,
+              private _toasterService: ToasterService,
+              private _authService: AuthService,
+              private _loaderService: LoaderService) {
+    this.configSettings = this._configService.getConfiguration();
+    this.newRegRegForm = _fb.group({
+      name: new FormControl(null, Validators.required),
+      description: new FormControl(null),
+      code: new FormControl(null, Validators.required),
+      state: new FormControl(null, Validators.required),
```
kjacobsen16: Can we change this to region to keep consistent with terminology?

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,437 @@
...
+      this._loaderService.showFullPageLoad();
+      this.addRegReg = false;
+      this._settingsService.getEntities(this.configSettings.regRegionURL + '/' + rr + '?includeGeometry=true')
+        .subscribe((res) => {
+          this.selectedRegRegion = res;
+          // set values in add regression region modal
+          this.newRegRegForm.controls['name'].setValue(this.selectedRegRegion.name);
+          this.newRegRegForm.controls['description'].setValue(this.selectedRegRegion.description);
+          this.newRegRegForm.controls['code'].setValue(this.selectedRegRegion.code);
+          this.newRegRegForm.controls['location'].setValue(this.selectedRegRegion.location);
```
kjacobsen16, line 227: I don't love all this, but I think I'll make an enhancement issue for cleaning up more.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,437 @@
...
+      this.configSettings.citationURL)
+      .subscribe((response: any) => {
+        this.newCitForm.reset();
+        this.addCitation = false;
+        rr.citationID = response.id;
+        if (!response.headers) {
+          this._toasterService.pop('info', 'Info', 'Citation was added');
+        } else { this._settingsService.outputWimMessages(response); }
+        this.cancelCreateRegression();
+        this._nssService.setSelectedRegion(this.selectedRegion);
```
kjacobsen16: I think this should be `requeryFilters` instead

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -0,0 +1,437 @@
...
+    if (this.polygonLayer) {
+      this.map.removeLayer(this.polygonLayer);
+      this.map.setView(new L.LatLng(39.8283, -98.5795), 4);
+    }
+    setInterval(() => {
+      this.map.invalidateSize();
+    }, 100);
+  }
+
+  // ADD COMMENTS
```
kjacobsen16: Should this be addressed before merging?

src/app/shared/components/addregressionregion/addregressionregion.component.html
```diff
@@ -0,0 +1,102 @@
...
+              <select class="form-control" formControlName="state">
+                  <option *ngFor="let reg of regions" [value]="reg.id">{{ reg.name }}</option>
+              </select>
+              <div class="input-invalid-warning" *ngIf="!newRegRegForm.get('state').valid && newRegRegForm.get('state').dirty">
+                  Region is required
+              </div>
+          </div>
+          
+          <button *ngIf="!uploadPolygon" type="button" class="btn-blue" (click)="showMap()">Upload Polygon</button>
+          <button *ngIf="!addCitation" type="button" class="btn-blue" (click)="addCitation = true" style="margin-left: 5px;">Add Citation</button>
```
kjacobsen16: My one comment here is that the flow is a little bit weird here, and I think moving those buttons would help. As it is now, the "Upload Polygon" button is above the citation form and before the "Add Citation" button if there's no citation. If you click it, the map shows up at the bottom. I think if you move these buttons to between the html for the new citation form and the map, with the upload polygon button after the add citation button, then the buttons would still be in line with each other but then the forms would show up closer to where the buttons are. Hope that makes sense.

Here's a recordit to show the workflow: https://recordit.co/PabQH96W6f

kjacobsen16 (changes requested):
Alright this is working pretty well! I did notice that if I'm playing around with it too much (loading too many regression regions) the zooming is weird, but I don't know how much that will affect the actual workflow. If you can find an easy fix for that, feel free. Otherwise the polygons are still showing up, it's just zoomed out so I think it's fine.

Great job - the functionality is there, my comments are mostly just some code cleanup.

kjacobsen16 (approved):
👍 I think we are good to go!

PR #192, opened 2020-05-13
Fixes #89

### LINKED ISSUE #89: Add Manage Citations button near edit scenario button

harper-wavra:


src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -2,25 +2,42 @@
...
-                        <div>{{l.title}}</div>
-                    </a>
-                    <br />
-                </div>
-            </div> 
-        </div>
-        <div class="modal-footer">
-            <button type="button" (click)="d('closed')" class="btn-black">Close</button>
-        </div>
+    <div style="padding-left: 20px; padding-right: 20px; padding-bottom: 20px;">
```
kjacobsen16: Generally, it's best not to have in-line styling. Try pulling these out into a class.

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -2,25 +2,42 @@
...
+    <input type="text" #input placeholder="Search" (keyup)="filter(input.value)">
+    <div style="padding-top: 20px;">
```
kjacobsen16: Same as above, should be a class

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -2,25 +2,42 @@
     <div class="modal-header">
         <div class="col-xs-12">
             <h4 class="modal-title pull-left">Manage Citations </h4>
-            <button class="close pull-right" (click)="d('closed')">x</button>
+            <button class="close pull-right" (click)="d('Closed')">x</button>
         </div>
     </div>
-    <div class="modal-body">
```
kjacobsen16, line 8: I'm not sure why this was removed, but I believe the `modal-body` class has some padding, so this would fix some of the comments below.

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -2,25 +2,42 @@
...
-                        <div>{{l.author}}</div>
-                        <div>{{l.title}}</div>
-                    </a>
-                    <br />
-                </div>
-            </div> 
-        </div>
-        <div class="modal-footer">
-            <button type="button" (click)="d('closed')" class="btn-black">Close</button>
-        </div>
```
kjacobsen16, line 24: I would put this back in to keep the format of our modals the same.

src/app/shared/components/managecitations/managecitations.component.css
```diff
@@ -11,6 +11,10 @@
     text-align: center;
     display: block;
 }
+.disabled-button{
```
kjacobsen16: I wouldn't call this `disabled-button` since you set it up for an `<i>` tag, not a `<button>`. Maybe just `disabled` or `disabled-icon`.

src/app/shared/components/managecitations/managecitations.component.css
```diff
@@ -11,6 +11,10 @@
...
+    opacity: 0.5;
+    pointer-events: none;
```
kjacobsen16: I would change this to `cursor: not-allowed !important;`. That will show the no-entry sign when you hover over it.

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -2,25 +2,42 @@
...
+                        <div>{{e.name}}</div>
+                    </span>
+                </td>
+                <td>
+                    <div *ngIf="!l.regressionRegions?.length>0">
+                        <i title="Delete Citation" (click)="deleteCitation(l.id)" class="fa fa-times" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
+                    </div>
+                    <div *ngIf="l.regressionRegions?.length>0">
+                        <i title="Delete Citation" (click)="deleteCitation(l.id)" class="fa fa-times disabled-button" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
+                    </div>
```
kjacobsen16: Instead of repeating things, you can have one `<i>` tag, but make certain properties conditional. In this case, you can set the click using `(click)="!l.regressionRegions?.length>0 && deleteCitation(l.id)"`.

You can also set conditional items using square brackets. For this item, you can do `[class.disabled-button]="l.regressionRegions?.length>0", which will only assign the disabled class if that statement is true.

Another way to set conditional statements is using the ternary/conditional operator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator. This is not applicable to this part of the code, but is definitely useful for setting conditions in the html or js. I use it a lot.

src/app/shared/components/managecitations/managecitations.component.ts
```diff
@@ -103,6 +113,23 @@ export class ManageCitationsModal implements OnInit, OnDestroy {
         this._settingsService.getEntities(this.configSettings.citationURL)
             .subscribe(res => {
                 this.citations = res;
+                this.filteredData = this.citations;
+                if(this.filterText){
```
kjacobsen16: This is a minor thing, so no need to change it but I usually try to put spaces after if statements and before `{`, so it would be `if (this.filterText) {`. I try to keep the code formatting consistent.

src/app/shared/components/managecitations/managecitations.component.ts
```diff
@@ -64,6 +70,18 @@ export class ManageCitationsModal implements OnInit, OnDestroy {
         this.modalElement = this.manageCitationsModal;
     }
 
+    public filter(input:string){
+        this.filteredData = this.citations.filter((item)=>{
+            let title = item.title.toLowerCase();
+            input = input.toLowerCase();
+            this.filterText = input;
+            if(title.includes(input))
```
kjacobsen16: We should check the author name as well

kjacobsen16 (changes requested):
I left a few comments on code changes. Most of them are small. I do think we should include the citation url somehow, maybe just do something similar to the mainview component, where the citation title/author text is a clickable link that opens in a new tab.

Otherwise I think this works well for now, but I'm going to make issues for some enhancements to it, both of which rely partially on changes coming through with issue 102 (still awaiting on re-projecting polygons sorry!)

kjacobsen16:
I'm also rethinking the columns now - "Title and Author" should maybe just be something like "Citation" or "Citation Info". And we should somehow make the title and authors more visibly separate. Maybe play with some bolding or italics? Or pulling out to separate columns?

harper-wavra:
I put the authors in italics, I thought the bold was kind of a lot to look at. Let me know if you think something different would look better.

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -5,22 +5,39 @@ <h4 class="modal-title pull-left">Manage Citations </h4>
...
     </div>
-    <div class="modal-body">
-        <h2 style="margin:0;padding:0">Citations</h2>
-        <div *ngFor="let sc of scenarios">
-            <div *ngIf="sc.citations">
-                <label *ngIf="sc.citations">{{sc.statisticGroupName}} Citations</label>
-                <div *ngFor="let l of sc.citations">
-                    <a *ngIf="l" [href]="l.citationURL" [target]="_blank">
-                        <div>{{l.author}}</div>
+    <div class="modalbody; table">
```
kjacobsen16: The class name should be `modal-body` with a hyphen, then you can probably remove the table class. Also, usually the way multiple classes are applied is by putting a space with no semi-colon.

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -5,22 +5,39 @@ <h4 class="modal-title pull-left">Manage Citations </h4>
...
-        </div>
+                </td>
+                <td>
+                    <span *ngFor="let e of l.regressionRegions; let eIndex = index">
+                        <div>{{e.name}}</div>
+                    </span>
+                </td>
+                <td>
+                    <div>
+                        <i title="Delete Citation" (click)="deleteCitation(l.id)" class="fa fa-times" [class.disabled-icon]="l.regressionRegions?.length>0" aria-hidden="true" style="cursor:pointer"></i>
```
kjacobsen16: This is missing the conditional portion I mentioned before. You can still click delete if the icon has the disabled class since it's just affected the css. The (click) should say something like `(click)="!l.regressionRegions?.length>0 && deleteCitation(l.id)"` so it'll only run the click function if there are no regression regions.

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -5,22 +5,39 @@ <h4 class="modal-title pull-left">Manage Citations </h4>
...
+                <th class="col-xs-5">Citation</th>
+                <th class="col-xs-5">Associated Regression Regions</th>
+                <th class="col-xs-1"></th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr *ngFor="let l of filteredData">
+                <td>
+                    <a [href]="l.citationURL" [target]="_blank">
                         <div>{{l.title}}</div>
```
kjacobsen16, line 23: I missed this before, but in the mainview the author is listed before the title, maybe we should do it that way to keep it consistent.

src/app/shared/components/managecitations/managecitations.component.ts
```diff
@@ -64,6 +70,19 @@ export class ManageCitationsModal implements OnInit, OnDestroy {
         this.modalElement = this.manageCitationsModal;
     }
 
+    public filter(input:string) {
+        this.filteredData = this.citations.filter((item)=>{
+            let title = item.title.toLowerCase();
+            let author = item.author.toLowerCase();
```
kjacobsen16: Can we also check the regression regions? It'll be a little trickier, but you can still have it on one line as I suggested above by using another filter on the regression regions and checking the length of what's returned. Try to get this working for a bit. If you get stuck, let me know and I'll point you in the right direction.

src/app/shared/components/managecitations/managecitations.component.ts
```diff
@@ -64,6 +70,19 @@ export class ManageCitationsModal implements OnInit, OnDestroy {
         this.modalElement = this.manageCitationsModal;
     }
 
+    public filter(input:string) {
+        this.filteredData = this.citations.filter((item)=>{
```
kjacobsen16: I must've rushed this the first time. Filters are nice in that you can just put the conditions in there without needing all the fluff. You can do this in one line. Here's a much simpler way to do this:
```
this.filteredData = this.citations.filter((item)=>
            item.author.toLowerCase().includes(input.toLowerCase()) || 
            item.title.toLowerCase().includes(input.toLowerCase()));
```

kjacobsen16 (changes requested):
I think I reviewed too quickly before and missed a couple edits - sorry about that! Again, mostly just some code enhancements.

kjacobsen16 (approved):
Looks great!

PR #198, opened 2020-05-18
Fixes #191

### LINKED ISSUE #191: Loader screen never closes when clicking "Compute"
If you click "Compute" on the sidebar and have left one or more values  blank the loader screen never closes.
image

From Pete's feedback 5/13/2020

harper-wavra:
Toggles the "Compute" on and off when "Edit Scenario" and "Finish Editing" are pushed. Also hides the loader when there are no values entered.

### DIFF
--- a/src/app/mainview/mainview.component.ts
+++ b/src/app/mainview/mainview.component.ts
@@ -1207,11 +1207,13 @@ export class MainviewComponent implements OnInit, OnDestroy {
 
     /////////////////////// Edit Scenarios Section ///////////////////////////
     public editRegScenario() {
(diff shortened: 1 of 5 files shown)


kjacobsen16 (commented):
My only comment is that having a no-entry sign on the cursor on this would be helpful. You should be able to edit the css file by doing something like:
```
button:disabled {
    cursor: not-allowed;
}
```
The reason it's written like that is because buttons have a `disabled` attribute, so you're saying if the button is disabled, add this property.

Otherwise works great!

kjacobsen16 (approved):
Looks great!

PR #200, opened 2020-05-21
Fixes #180

### LINKED ISSUE #180: Add filtering to dropdown
Filter the drop downs based on user text input that searches both the name and the code (or whichever is shown). There are probably libraries or functions that can be set up.

From Pete's Feedback 4/16/20

NathanKrawza:
Uses ng-select. Adds search/filtering for drop downs specifically in addscenario.

### DIFF
--- a/src/app/shared/components/addscenario/addscenario.component.html
+++ b/src/app/shared/components/addscenario/addscenario.component.html
@@ -24,11 +24,9 @@
                 <label class="req" for="statisticGroupID">Statistic Group:</label>
                 <ng-select 
                 formControlName="statisticGroupID"
-                [items]="statisticGroups"
                 [selectOnTab]="true"
-                [closeOnSelect]="true"
-                bindValue="id"
-                bindLabel="name">
+                [closeOnSelect]="true">
+                <ng-option *ngFor="let sg of statisticGroups" [value]="sg.id">{{sg.name}}</ng-option>
                 </ng-select>
                 <!-- <select class="form-control" formControlName="statisticGroupID">
                     <option *ngFor="let sg of statisticGroups" [value]="sg.id">{{sg.name}}</option>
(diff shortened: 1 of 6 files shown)


src/app/shared/components/addscenario/addscenario.component.html
```diff
@@ -22,9 +22,9 @@ <h4 class="modal-title pull-left" *ngIf="clone==false"> New Scenario </h4>
             <!-- Statistic Group* -->
             <div class="form-group required" [ngClass]="{'form-invalid': !newScenForm.get('statisticGroupID').valid && newScenForm.get('statisticGroupID').dirty}">
                 <label class="req" for="statisticGroupID">Statistic Group:</label>
-                <select class="form-control" formControlName="statisticGroupID">
-                    <option *ngFor="let sg of statisticGroups" [value]="sg.id">{{sg.name}}</option>
-                </select>
+                <ng-select formControlName="statisticGroupID" [selectOnTab]="true" [closeOnSelect]="true" [clearable]="false">
+                <ng-option *ngFor="let sg of statisticGroups" [value]="sg.id">{{sg.name}}</ng-option>
```
kjacobsen16: Tab these for readability.

kjacobsen16 (changes requested):
Looks and works great for adding a scenario. However, cloning a scenario has issues - my guess is it has to do with how the existing values are sent through compared to how the ng-select reads it, similar to how we had to remove the `toString()` in a few lines. The IDs should be integers.

kjacobsen16:
@harper-wavra Nathan is off to his new internship, can you try to fix this? There's an issue with the unit selects when cloning a scenario, I'm guessing they just need to be set differently in the cloneScenario() function. Figured you'd be good for this since you set up the cloneScenario() function :) It may help to look at how the units are set when creating a new scenario v. in the clone scenario function.

src/app/shared/components/addscenario/addscenario.component.html
```diff
@@ -37,19 +37,18 @@ <h4 class="modal-title pull-left" *ngIf="clone==false"> New Scenario </h4>
                         <label class="req" for="ID">Regression Region:</label>
                         <i *ngIf="loggedInRole == 'Administrator'" (click)="showAddRegRegion()" title="Add Regression Region" class="ion-plus-circled"></i>
                     </span>
-                    <select class="form-control" formControlName="ID">
-                        <option selected hidden value=""></option>
-                        <option *ngFor="let rr of regressionRegions" [value]="rr.id">{{rr.name}}</option>
-                    </select>
+                    <ng-select formControlName="ID" [selectOnTab]="true" [closeOnSelect]="true" [clearable]="false">
+                    <ng-option *ngFor="let rr of regressionRegions" [value]="rr.id">{{rr.name}}</ng-option>
```
kjacobsen16: Whoops, this was missed. This should be tabbed

kjacobsen16 (approved):
Works great! 👍

PR #202, opened 2020-05-21
fixes #187

### LINKED ISSUE #187: Add "Get Started" section to About modal
Put above last sentence? Something like:
```
<h2> Get Started! </h2>
<p>Navigate to the sidebar to explore the available regression-based scenarios based on region. Filter to scenarios of interest, fill in the necessary parameter values on the main screen, and click "Compute" on the sidebar to view results and citations. Click on "?" icons for additional help.</p>
```

NathanKrawza:


kjacobsen16 (changes requested):
While we're at it, I noticed a couple changes we should make: there's a typo in the second paragraph. `StreamStats also provide access` should be `StreamStats also provides access`.

Also, can you make those about headers bigger? The whole sheet has `font-size: 98% !important;` applied to it. I'm not sure why, but I wouldn't remove it since it'll affect everything else. Either way, the h2 tags are really small because of it.

kjacobsen16 (approved):
👍

PR #203, opened 2020-05-21
Fixes #188

### LINKED ISSUE #188: Make citation link clickable
image

Anywhere we have a link, we should make it into an anchor tag, with target="_blank"

harper-wavra:
Made citation url clickable. I looked around for other links but didn't see any, let me know if you know of any more.


kjacobsen16 (approved):
Looks good

PR #208, opened 2020-06-02
Fixes #190

### LINKED ISSUE #190: Change StatLabel to Definition in parameter drop down
When creating a new scenario in the parameter drop down it displays the name of the parameter and the StatLabel, example: "Base Flow Index (BFI)." Instead of the StatLabel, we want to display the definition of the parameter. 

From Ryan's Feedback

harper-wavra:
When creating a new scenario changed parameter label code to description


kjacobsen16 (approved):
I figured we'd need the code shown, so I just went in and changed it quick to code (description). I checked with Ryan and he likes that idea.

PR #211, opened 2020-06-05
Kj patch

### LINKED ISSUE #71: Part of Todd's review - Issue 1, weird rounding for AK D15
image
Should be 11,800,000
The value coming from nssservices is 11812235.2250751

Other equations tend to round correctly.

kjacobsen16:
- fixes #71 by removing client-side rounding (coming from services now)
- skip over duplicate messages from server to avoid duplicate toasts
- rebuild


harper-wavra (approved):
Worked for me

PR #214, opened 2020-06-08
Fixes #195

### LINKED ISSUE #195: "Refresh" when adding new Regression Region
When you add a "new Regression Region" from the "New Scenario" modal the regression region doesn't show up unless you close and reopen the modal. We need a way to refresh the Regression Region list in the modal without opening and re-closing the modal.

harper-wavra:
Removed `getRegRegions()` in `addscenario `and `addregressionregion `component and used regression region subscriber to updated regression region list.


kjacobsen16 (approved):
Works great!

PR #215, opened 2020-06-11
Fixes #162

### LINKED ISSUE #162: Add a citation manager in the Edit/Add Regression Region modal.
We need a way to assign existing citations to regression regions when adding or editing. This could be in the form of a button that says "Find existing citation" or something like that, which would allow the user to filter through the citations using a text input and select a citation to assign.

This will need to wait until #102 is merged into dev

harper-wavra:
Able to assign existing citations when adding/editing regression regions.

### DIFF
--- a/src/app/shared/components/addregressionregion/addregressionregion.component.ts
+++ b/src/app/shared/components/addregressionregion/addregressionregion.component.ts
@@ -54,7 +54,6 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
   private polygonLayer;
   private selectedCitation;
   public currentCitation;
-  
   public selectedRegressionRegion: Array<Regressionregion>;
   public tempSelectedRegressionRegion: Array<Regressionregion>;
 
@@ -120,7 +119,7 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
     });
     this._nssService.currentCitation.subscribe(item => {
       this.currentCitation = item
-      if (this.currentCitation) {
+      if (this.currentCitation != " ") {
         this.addExistingCitation();
       }
     });
@@ -199,9 +198,10 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
   }
 
   public addExistingCitation(){
+    this.newRegRegForm.controls['citationID'].setValue(this.currentCitation.id);
     this.newCitForm.controls['title'].setValue(this.currentCitation.title);
     this.newCitForm.controls['author'].setValue(this.currentCitation.author);
-    this.newCitForm.controls['citationURL'].setValue(this.currentCitation.citationURL);
+    this.newCitForm.controls['citationURL'].setValue(this.currentCitation.citationURL);  
   }
   
   outputWimMessages(msg) {
@@ -315,7 +315,13 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
           this._toasterService.pop('info', 'Info', 'Regression region was added');
         } else { this._settingsService.outputWimMessages(response); }
         if (this.addCitation) { // if user elected to add a citation, send that through
-          this.createNewCitation(response);
+          if (this.newRegRegForm.value.citationID != this.currentCitation.id || !this.newRegRegForm.value.citationID) {
+            this.createNewCitation(this.selectedRegRegion);
+          } else {
+            this._loaderService.hideFullPageLoad();
+            this.requeryFilters();
+            this.cancelCreateRegression();
+          }
         } else {
           this.cancelCreateRegression();
           this.requeryFilters();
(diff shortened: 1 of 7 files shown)


harper-wavra:
I think this is working now. You are able to assign existing citations when adding/editing regression regions. I fixed the issue with editing an existing regression region, removing the citation, then adding an existing citation. Also set the `citationID` to `currentCitation.id` so it is no longer creating a new citation each time. 

When using an existing citation I made the input read only so a citation would not accidentally get edited for all of its associated regression regions. I think it would be better to have the "edit citation" option inside the mange citation modal that way it is clear when you edit that citation it will be edited for all the regression regions that use that citation instead of just the one you are working on in the edit/add regression region modal. Let me know what you think.

kjacobsen16:
I think that's a great idea! Go ahead and add an issue for that. If you do that, you can also remove the edit button next to the citations in the main edit scenario page I think.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -23,6 +23,7 @@ import { LoaderService } from 'app/shared/services/loader.service';
 import { Regressionregion } from 'app/shared/interfaces/regressionregion';
 import { Statisticgroup } from 'app/shared/interfaces/statisticgroup';
 import { Regressiontype } from 'app/shared/interfaces/regressiontype';
+import { __core_private_testing_placeholder__ } from '@angular/core/testing';
```
kjacobsen16: Looks like an accidental import

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -100,6 +103,7 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
     });
     this.modalSubscript = this._nssService.showAddRegRegionModal.subscribe((result: AddRegressionRegion) => {
       if (result.show) { 
+          this.rr = result.regRegionID
```
kjacobsen16: tiny issue but this is missing a `;`

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -117,6 +121,12 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
     this._nssService.selectedRegRegions.subscribe((regRegions: Array<Regressionregion>) => {
       this.selectedRegressionRegion = regRegions;
     });
+    this._nssService.currentCitation.subscribe(item => {
+      this.currentCitation = item
```
kjacobsen16: Also missing `;`

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -297,7 +319,13 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
         } else { this._settingsService.outputWimMessages(response); }
         if (this.addCitation) { // if user elected to add a citation, send that through
-          this.createNewCitation(response);
+          if (this.newCitation == true) {
+            this.createNewCitation(this.selectedRegRegion);
+          } else {
+            this._loaderService.hideFullPageLoad();
+            this.requeryFilters();
+            this.cancelCreateRegression();
+          }
```
kjacobsen16: One of these `if`s is unnecessary. You could just do `if (this.addCitation && this.newCitation == true)` on line 321 and cut lines 322 and 324-327, since adding an existing citation just involves editing the citationID, not creating a new citation. That way it'll just do the regular stuff (cancelCreateRegression, requeryFilters) if you're not creating a _new_ citation.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -310,6 +338,78 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
+                });
+            } else if (this.addCitation) { // New citation
+             
+              if (this.newCitation == true) {
+                this.createNewCitation(this.selectedRegRegion);
+              } else {
+                this._loaderService.hideFullPageLoad();
+                this.requeryFilters();
+                this.cancelCreateRegression();
+              }
```
kjacobsen16: Same here, this can be shortened by doing 
```
} else if (this.addCitation && this.newCitation == true) {
  this.createNewCitation(this.selectedRegRegion);
}
```

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -310,6 +338,78 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
+            }
+        }, error => {
+            this._loaderService.hideFullPageLoad();
+            if (this._settingsService.outputWimMessages(error)) {return; }
+            this._toasterService.pop('error', 'Error editing Regression Region', error.message || error._body.message || error.statusText); }
+        );  
+  }
+
+  public removeCitation(){
+    this.saveFilters();
```
kjacobsen16, line 379: The tabbing in this function is off

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -310,6 +338,78 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
+            }, error => {
+                if (this._settingsService.outputWimMessages(error)) {return; }
+                this._toasterService.pop('error', 'Error removing Citation', error._body.message || error.statusText);
+            }
+        );
+        this.selectedRegRegion.citationID=null;
+        this.newRegRegForm.controls['citationID'].setValue(null);
+        this.newCitForm.controls['title'].setValue(null);
+        this.newCitForm.controls['author'].setValue(null);
+        this.newCitForm.controls['citationURL'].setValue(null);
```
kjacobsen16: Resetting the `newCitForm` (line 410) should reset all the controls in it, so this isn't needed.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -310,6 +338,78 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
+      
+      this.addCitation = false; 
+      this.newCitForm.reset(); 
+      this.newCitation = true
```
kjacobsen16: Missing `;`

src/app/shared/services/app.service.ts
```diff
@@ -101,6 +101,20 @@ export class NSSService {
     public get showManageCitationsModal(): any {
         return this._showHideManageCitationsModal.asObservable();
     }
+    // Toggle Add/Delete Citation in managecitations modal
+    private addCitations = new BehaviorSubject<boolean>(true);
+    currentAddCitations = this.addCitations.asObservable();
+
+    showAddCitations(bool: boolean){
+        this.addCitations.next(bool);
+    }
```
kjacobsen16: This all works well, but could probably be combined with the showManageCitations modal stuff, similarly to how we're sending things along to the showHideAddRegressionRegion modal, so you're only setting one subject when you want the modal shown.

kjacobsen16 (changes requested):
It's working well, I just made some code cleanup/improvement suggestions.

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -310,6 +335,66 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
+                    this._settingsService.outputWimMessages(response); 
+                    this.cancelCreateRegression(); 
+                  }
+                  this.requeryFilters();
+                  this.cancelCreateRegression();
+                }, error => {
+                  this._loaderService.hideFullPageLoad();
+                  if (this._settingsService.outputWimMessages(error)) { return; }
+                  this._toasterService.pop('error', 'Error creating Citation', error.message || error._body.message || error.statusText);
+                });
```
kjacobsen16, line 364: If we're not editing the citation properties in the regression region component, then this won't be needed anymore. We can leave it in for now while we're waiting to get the new method working, but something to think about.

kjacobsen16 (approved):
Working well!

PR #217, opened 2020-06-15
Fixes #193

### LINKED ISSUE #193: Limit bias correction factor to admin users
Bias correction factor inputs in the add scenario modal and edit scenarios page (part of the prediction interval), should be disabled for non-admin users, similarly to how we only allow admins to view the settings page.

From Pete's review

harper-wavra:
Removes Bias Correction Input when logged in as Manager. Also has some fixes (checks for null) and temporary fixes (remove auth header) for Manage Citation modal when logged in as Manager.

### DIFF
--- a/src/app/shared/components/managecitations/managecitations.component.ts
+++ b/src/app/shared/components/managecitations/managecitations.component.ts
@@ -83,7 +83,7 @@ export class ManageCitationsModal implements OnInit, OnDestroy {
 
     public filter(input:string) {
         this.filterText = input;
-        this.filteredData = this.filteredData.filter(c => 
+        this.filteredData = this.citations.filter(c => 
             c != null &&
             (c.author.toLowerCase().includes(input.toLowerCase()) ||
             c.title.toLowerCase().includes(input.toLowerCase()) ||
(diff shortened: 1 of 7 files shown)


src/app/shared/components/managecitations/managecitations.component.ts
```diff
@@ -125,10 +128,12 @@ export class ManageCitationsModal implements OnInit, OnDestroy {
...
 
     public getCitations() {
-        this._settingsService.getEntities(this.configSettings.citationURL)
+        this._settingsService.getEntitiesCitations(this.configSettings.citationURL)
             .subscribe(res => {
                 this.citations = res;
-                this.filteredData = this.citations;
+                this.filteredData = this.citations.filter(function (filter) {
+                    return filter != null;
+                });
```
kjacobsen16: I wonder if it's better here to, instead of creating a new item in the settings service just for this one call, just make an http call here instead. It works either way, but might be cleaner.

kjacobsen16 (changes requested):
We'll also want to limited the editing of the bias correction factor in the "Edit scenario" page to admins. There are a few regressions in the Test KJ reg region that have prediction intervals/bias correction factors.

kjacobsen16 (approved):
👍

PR #219, opened 2020-06-22
Fixes #157

### LINKED ISSUE #157: Display Regression Regions with no Scenarios during "Edit Scenarios"
When you click the "Edit Scenarios" button, only Regression Regions that have Scenarios are displayed. We should display Regression Regions that have no Scenarios.

harper-wavra:
Added function to get unused regression regions, then displays them in the edit scenario.

### DIFF
--- a/src/app/mainview/mainview.component.ts
+++ b/src/app/mainview/mainview.component.ts
@@ -123,6 +123,8 @@ export class MainviewComponent implements OnInit {
     public addRegReg: boolean;
     public selectedRegRegion;
     public modalRef;
+    public regRegionsScenarios = [];
+    public unusedRegRegions;
     // public changeStatGroup = false;
     public citations: Array<Citation>;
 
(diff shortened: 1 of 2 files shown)


src/app/mainview/mainview.component.ts
```diff
@@ -1389,6 +1391,19 @@ export class MainviewComponent implements OnInit {
...
+        this.regressionRegions.forEach(x=> {
+            this.scenarios.forEach(s=> {
+                s.regressionRegions.forEach(rr => {
+                    if (x.id == rr.id) {
+                        this.regRegionsScenarios.push(x)
+                    }
+                });
+            });
+        });
+        this.unusedRegRegions = this.regressionRegions.filter(entry1 => !this.regRegionsScenarios.some(entry2 => entry1.id === entry2.id));
```
kjacobsen16, line 1405: This could probably slimmed down a bit using nested filters, but it works so I'm ok with it 👍

kjacobsen16 (changes requested):
Works well, the only issue I'm noticing is when I removed the regressions from my reg region (so there were now no scenarios for that region), it didn't show it in the list until I refreshed. So, I think we just need to update the `unusedRegRegions` anytime the scenarios change.

kjacobsen16 (approved):

PR #222, opened 2020-07-02
Fixes #220

### LINKED ISSUE #220: Issues with "removing" citation from new regression region modal
If I go to create a new regression region, hit "add new citation" then try removing it, I get an error.

Additionally, if I remove a citation then hit "Save", the citation isn't being completely removed. It's removed until I hit save, then assigned back to the regression region. I'm guessing the citation ID isn't being overwritten properly.

Maybe we shouldn't remove the citation before they hit save like it's currently doing - I know it messes up the list of citations in the manager citations modal, but the user should understand that they haven't saved yet so it's still assigned.

harper-wavra:
Solves issue #220 as well as issue when cloning a scenario and choosing a new region the regression regions drop down list didn't update for the new region selected.

### DIFF
--- a/src/app/shared/components/addregressionregion/addregressionregion.component.ts
+++ b/src/app/shared/components/addregressionregion/addregressionregion.component.ts
@@ -397,9 +397,7 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
   }
 
   public removeCitation(){
-    if (this.selectedRegRegion) { // if creating new regression region cannot set citationID to null
-      this.selectedRegRegion.citationID = null
-    }
+    this.newRegRegForm.controls['citationID'].setValue(null);
     this.removeCit = true;
     this.addCitation = false; 
     this.newCitForm.reset(); 
--- a/src/app/shared/components/addregressionregion/addregressionregion.component.ts
+++ b/src/app/shared/components/addregressionregion/addregressionregion.component.ts
@@ -57,8 +57,9 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
   public currentCitation;
   public selectedRegressionRegion: Array<Regressionregion>;
   public tempSelectedRegressionRegion: Array<Regressionregion>;
-  public newCitation: boolean;
+  public newCitation: boolean = false;
   public rr;
+  public removeCit: boolean;
 
   public tempSelectedStatisticGrp: Array<Statisticgroup>;
   public get selectedStatisticGrp(): Array<Statisticgroup> {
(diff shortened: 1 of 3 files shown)


src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -339,60 +340,70 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
+              this.requeryFilters();
+              this._nssService.outputWimMessages(response);
+            }, error => {
+              if (this._settingsService.outputWimMessages(error)) {return; }
+              this._toasterService.pop('error', 'Error removing Citation', error._body.message || error.statusText);
+            }
+          );
+          this.selectedRegRegion.citationID = null;
+          this.cancelCreateRegression();
+        }
```
kjacobsen16: Instead of doing a whole other `putEntity` call here, your `removeCitation` function should be changing the citation ID to null for the reg region, which will go through the putEntity call used to update the reg region. In that case, this whole section can be removed.

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -198,6 +198,12 @@ export class AddScenarioModal implements OnInit, OnDestroy {
         this.newScenForm.get('region').valueChanges.subscribe(item => {
              if(item != null){
                 this.selectedRegion = item;
```
kjacobsen16: If we change this to `this._nssService.setSelectedRegion(item);`, it'll update the regression regions _and_ the selected region, which means we won't need lines 202-205

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -339,60 +340,70 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
-            this.requeryFilters();
-            this._nssService.outputWimMessages(response);
-        }, error => {
-            if (this._settingsService.outputWimMessages(error)) {return; }
-            this._toasterService.pop('error', 'Error removing Citation', error._body.message || error.statusText);
-        }
-    );
-    this.selectedRegRegion.citationID=null;
+    if (this.selectedRegRegion) { // if creating new regression region cannot set citationID to null
+      this.selectedRegRegion.citationID = null
```
kjacobsen16, line 383: This needs to change the citation ID of the newRegReg form. Otherwise, it doesn't get sent through when saving the reg region.

kjacobsen16 (changes requested):
In general, things are working but there are a few places where you're making additional calls that are unnecessary. Make sure to check that you're not replicating functions and calls, especially when it comes to things that we can set up subscribers for (e.g. selected region, reg regions, etc.).

kjacobsen16 (approved):
Everything appears to be working great!

PR #226, opened 2020-07-08
Styles

mitchas:
Main changes:

- Responsive adjustments to text/button sizes on mobile

- Mobile menu button and pop-out menu fixed on mobile (<700px width)

- Responsive changes to modals, aligned at bottom of screen, full width, better resizing


harper-wavra (commented):
It looks really good!! Just a couple things I noticed.

1. In the Manage Citation modal the delete citation and add citation button is cut off.
[image](https://user-images.githubusercontent.com/59063181/87066515-50bece00-c1d8-11ea-9cb5-6672683fc6c9.png)

2. The buttons to add a new Regression Region are gone.
[image](https://user-images.githubusercontent.com/59063181/87066686-91b6e280-c1d8-11ea-802a-f4f741ab4493.png)
[image](https://user-images.githubusercontent.com/59063181/87068252-e8252080-c1da-11ea-92ef-aa548d8a355a.png)

3. When editing a regression region the 'Add New Citation", "Find Existing Citation", and "Upload Polygon" buttons don't really look like buttons. Not sure if this was on purpose or not.
[image](https://user-images.githubusercontent.com/59063181/87067713-0cccc880-c1da-11ea-8550-a0a27163bbf0.png)

4. Using mobile the setting pages and the navigational bar on top didn't let me scroll over. Not sure if I tested mobile the right way though so this might not be an issue, or even if mobile will be used. I went into the developer tools and clicked on "toggle device toolbar."
[image](https://user-images.githubusercontent.com/59063181/87067820-34bc2c00-c1da-11ea-9ab6-3b75c3743ce9.png)

kjacobsen16:
Thanks @harper-wavra 

Mitch, this is looking nice! I also noticed

1. If you compute scenarios, the margins are gone.:
image
- To replicate, select a region in the main page, put in whatever numbers for the parameter values and hit "compute"

2. If you put in a parameter value outside the suggest range, the table changes size, just threw me off
image

3. In modals, such as the Add scenario modal, it's not very clear that you can scroll to see more. Also, in the past we had the body of the modal scrollable, but the footer was stationary.

4. In the Add Scenario modal, I had the parameter min and max on the same line before. I'm not sure which is better, but it does save some space
image

5. In the main page when logged in, if you click "Edit Scenario", you can no longer see the equations (they are mathjax)
image

6. The toast messages display over the nav links 
image

7. In the add/edit regression region modal, if you remove the citation and hit "Find existing citation" (which brings up an additional modal), and filter the citations (which makes the modal smaller), you can't tell which modal is which.
image

Let me know if you have questions about how to replicate :)

mitchas:
Alright, a more here than last time:

- All the issues from KJ and Harper above should be fixed.

- Bootstrap fully removed (hopefully)

- New table styles, and a bunch of options for tables (I'll have to demo on call sometime)

- All icons are back. Switched to the latest Font Awesome pro

- Added scroll bars to modals

- The tabs at the top of modals are now fixed, don't scroll

- Proper tooltips

- Added the new Favicon, and other meta/share images

-  Switched images to local references

harper-wavra (commented):
Hello - 

I noticed and I am not sure if I am just missing something but I am not sure how to log in now (there used to be a Manage button in the top navigation bar which I think is now gone.)  [image](https://user-images.githubusercontent.com/59063181/88069531-328d9200-cb37-11ea-974e-135a36bdd436.png)

kjacobsen16:
Looks like he must've pulled in staging - the "loginShow" option in the config.json is set to "false" https://github.com/USGS-WiM/NSS/blob/Styles/src/assets/config.json

harper-wavra (commented):
Once I changed `loginShow` to true I was able to login. I thought it looked very nice and didn't notice anything that wasn't working.

harper-wavra:
Looks good!

harper-wavra (approved):

PR #233, opened 2020-07-10
Fixes #216

### LINKED ISSUE #216: Move edit citation to manage citation modal
Move the "edit citation" option in the main edit scenario page to inside the mange citation modal. Can also remove the edit button next to the citations in the main edit scenario page.

harper-wavra:
Added edit citation button to manage citation modal, as well as removed the edit citation from the edit scenario section.

### DIFF
--- a/src/app/mainview/mainview.component.html
+++ b/src/app/mainview/mainview.component.html
@@ -719,8 +719,6 @@
                                 <tr *ngFor="let c of citations | citationfilter:rr.citationID; let cIndex = index;">
                                     <td>
                                         <div class="data-controls">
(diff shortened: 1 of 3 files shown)


kjacobsen16 (commented):
This works well, but I think it needs some styling help:
1. With longer titles, you have to scroll over to the right to see the buttons. Setting a max-width should help
2. Also with longer titles, you can't see the full names in the inputs - I wonder if `textarea` would be better.

@mitchas should I just merge this and let you fix it while you're already doing styling stuff?

kjacobsen16 (approved):

Issue #227, opened 2020-07-09
Add Gagestats to top navigation bar


harper-wavra:
Add the Gagestats button to the top navigation bar. We should also consider renaming NSS tab to something else maybe Regression Equations?

![mock-up](https://user-images.githubusercontent.com/59063181/87049947-31697600-c1c3-11ea-8329-4bc7d1037119.png)

ghost:
What should happen when that Gagestats button is clicked?

esmyers:
That's a good question @09Kathy I was also thinking about this so I'll chime in.  I was thinking that the GageStats button would activate a routerLink and the router-outlet would be the Mainview component which would load the GageStats component into the red box on screen.  This is sort of just my first pass though, I'm open to other ideas.
![image](https://user-images.githubusercontent.com/2495167/87166365-19602800-c291-11ea-89d3-e0eb95b701c7.png)

kjacobsen16:
Yes - you can follow what the nav link for the "Settings" page does. Also - we should have the NSS link first in the navbar, since I think the NSS page will be the default (I made this wireframe a while ago, sorry).

kjacobsen16:
Whoops - should've tagged you guys @09Kathy @esmyers

ghost:
I want to make sure I understand this correctly - will Gagestats, NSS, and Settings replace the current Home and About buttons? If no Regions, etc. are chosen would Gagestats just route back to the home screen (like what the Home button does right now?)
![image](https://user-images.githubusercontent.com/46943254/87172468-e9695280-c299-11ea-817d-246d663a8b5f.png)

ghost:
We can also just wait to go over this when we meet later today too!

kjacobsen16:
So, I did the nav bar a little wrong in my wireframe. We want "NSS" to replace the "Home" button, so the "NSS" page is the default when you open the application. Gagestats will be like the Settings page, an additional nav link that's a separate component, so you'll probably want a whole "GageStats" folder in the code similar to the settings. Also, the gagestats button should be to the right of the NSS button, ignore the order in my wireframe.

The about link should stay in the nav bar.

Also, there's more on this in issue #228 and #229, you can just get the general functionality with placeholder text set up, then we can fill in the page and sidebar with those issues.

PR #234, opened 2020-07-10
#227 fix

Anders-Hopkins:
Added Gagestats utton to navbar and created Gagestats component


kjacobsen16 (approved):
Looks great, you must've used the CLI!

We can't see your `gagestats works` text because of styling, but that's fine, we'll fix that in other issues.

PR #239, opened 2020-07-17
Lm 136

ltmetz:
Add agencies and station type to lookups on Settings page


dist/assets/config.json
```diff
@@ -1,5 +1,7 @@
 {
-   "baseURL": "https://test.streamstats.usgs.gov/nssservices/",
+   "nssBaseURL": "https://test.streamstats.usgs.gov/nssservices/",
+   "gageBaseURL": "https://test.streamstats.usgs.gov/gagestatsservices/",
```
kjacobsen16: I think to keep terminology consistent/easier to follow, we should call this this `gageStatsBaseURL` instead.

src/app/settings/categories/agencies/agencies.component.html
```diff
@@ -0,0 +1,89 @@
+<div>
+    <legend style="margin-bottom:25px;margin-top:0;">Agencies</legend>
+</div>
+<button [hidden]="loggedInRole !== 'Administrator'" type="button" class="btn-blue" (click)="showNewAgencyForm()"><i class="fa fa-plus"
```
kjacobsen16: This is probably copied from the other settings components, but since the settings component is only available for admins, we can remove this check for loggedinrole == admin (originally, we had the settings available for any logged in users)

src/app/settings/settings.component.html
```diff
@@ -7,10 +7,12 @@
         <div id="sidebar-wrapper">
             <h1>Lookups</h1>
             <ul class="settings-links">
+                <li><a [class.hidden]="loggedInRole !== 'Administrator'" [routerLink]="['agencies']" routerLinkActive="active">Agencies</a></li>
```
kjacobsen16: Same as above, we can actually remove these checks for the loggedinrole == admin since the settings pages are only accessible by

src/app/settings/settings.service.ts
```diff
@@ -109,11 +130,30 @@ export class SettingsService {
             .catch(this.errorHandler);
     }
 
+    public postEntityGage(entity: object, url: string) {
```
kjacobsen16: Same here, I wonder if we should name everything "GageStats" for consistency?

kjacobsen16 (approved):
Works well, marking as approved but there are some cleanup/text changes that would be good to do before merging (comments made).

Also - we'll have to warn people that there's a disconnect between the users in NSS v. GageStats DBs. If someone creates a new user or tries to login with a user that's in NSS but not gagestats, it won't work. We might have to add an issue to duplicate the manager add/edit calls to also update the gagestats user endpoint.

kjacobsen16 (approved):
Looks good!

PR #242, opened 2020-07-23
Fixes #238

### LINKED ISSUE #238: Incorporate Status for regression regions
You can get a list of available status here: https://test.streamstats.usgs.gov/docs/nssservices/#/Status/GET/Status.
- [x] Each regression region has a "StatusID" - the services filter scenarios and regression regions by status ID. We should show the status of each regression region, maybe in the edit scenario portion

- [x] Allow users to edit status in adding or editing regression region modal, default for new reg regions should be status ID = 2 (review)

harper-wavra:
Displays statusID in the edit scenario portion.
Allows users to edit status of regression region. Default of new regression region set to status ID of 2
Small fix for adding existing citations


src/app/mainview/mainview.component.ts
```diff
@@ -1405,6 +1412,16 @@ export class MainviewComponent implements OnInit {
...
     }
     
+    public getStatusDescription(sID) {
+        this.tempStatusID = sID;    
+        this.status.forEach(z => {
+            if (this.tempStatusID === z.id) {
+                this.tempStatusName = z.name;
+            }
+        });
+        return (this.tempStatusName);
```
kjacobsen16: If you're not using these variables elsewhere, you don't have to create global variables. You can use either a local variable or use a one line `find()` function and return the response from that. I'll post my guess at both for learning purposes (not testing it out so you may have to adapt).

e.g. with local variable :
```
let statusName;
this.status.forEach(z => {
            if (sID === z.id) {
                statusName = z.name;
            }
        });
return statusName;
```

or 1 line:
```
return this.Status.find(s => s.id == sID).Name;
```

kjacobsen16 (approved):
Looks good, only thing we should change is having the description in the dropdown display in the add/edit regression region modal, like `Review (description of review status)` because most users might not know what the status means.

Also left a comment for a code enhancement.

Otherwise, approved.

harper-wavra:
Merged dev to branch and resolved KJ's comments.

PR #243, opened 2020-07-24
Fixes 240, Duplicated manager add/edit calls to update gagestats user endpoint

ltmetz:


src/app/shared/components/profile/profile.component.ts
```diff
@@ -80,6 +82,10 @@ export class ProfileComponent implements OnInit, OnDestroy {
         this._settingsService.getEntities(this.configSettings.managersURL + '/' + this.loggedInID).subscribe(res => {
             this.userInfo = res;
         });
+        // TODO: Delete once users/managers tables are connected between databases.
+        this._settingsService.getEntitiesGageStats(this.configSettings.usersURL + '/' + this.loggedInID).subscribe(res => {
+            this.userInfoGageStats = res;
+        });
```
kjacobsen16: This doesn't work right, since the `loggedInID` is the ID from the NSS db, and is different than gagestats. Instead, you could take the username from the response above (lines 82-84) and find the user from the gagestats user list that matches that.

kjacobsen16 (changes requested):
Mostly working great. I left one note on an issue with the profile component's save function. It saved in NSS, but not in GageStats, so be sure to keep an eye on your console errors so you know when an error occurs.

Also, I wonder if we should have some sort of a flag to let the user know who's in gagestats DB, since some of the pre-existing users aren't in there yet. Maybe a column in the managers page for now (that we take out once the tables are merged) that's checked if they're in gagestats (so check to see if a user exists with that username in the gagestats users list)? Hope that makes sense!

kjacobsen16:
Closing, issue was ultimately fixed in the services, thanks for all the effort!

PR #244, opened 2020-07-27
Fixes 224

ltmetz:
Deleted a line of code that was removing form controls for regions, and looks to have fixed the problem/bug.


harper-wavra (approved):
Works well

PR #247, opened 2020-07-29
Fixes 228 and 229

Anders-Hopkins:
Added gage stats side bar with station type and agencies as parameters. A table is created based off of station type parameter


src/assets/config.json
```diff
@@ -4,13 +4,15 @@
     "agenciesURL": "Agencies",
     "citationURL": "Citations",
     "errorsURL": "Errors",
+    "gageURL": "https://test.streamstats.usgs.gov/gagestatsservices/",
```
kjacobsen16: Since the gageStatsBaseURL was added earlier, you can remove this line. You can do a full Find and Replace in VSCode.

src/app/app.component.html
```diff
@@ -56,13 +56,15 @@
 	</div>
 
 </nav>
-
+<!-- <wim-sidebar *ngIf="router.url.indexOf('/settings') === -1 && router.url.indexOf('/profile') === -1 && router.url.indexOf('/gagestats') === -1"></wim-sidebar> -->
+<gs-sidebar *ngIf="router.url.indexOf('/gagestats') > -1"></gs-sidebar>
```
kjacobsen16: To keep consistent with style changes, we should maybe move this to after line 63. That'll help with styling.

src/app/gagestats/gagestats.component.ts
```diff
@@ -8,18 +16,82 @@ import { GagestatsService } from './gagestats.service';
...
+      this.selectedAgency = Agency;
+      console.log(this.selectedAgency)
+      if (this.selectedAgency > 0) {
+        this.selectStations = [];
+        this._nssService.getStationsByAgency(this.selectedStationType.id, this.selectedAgency.id).subscribe((s: Array<Station>) => {
+          this.selectedStations = s;
+           });
+      }
+    });
+    */
```
kjacobsen16: Is this necessary to keep in? Maybe comment why you're keeping it if so?

src/app/gs-sidebar/gs-sidebar.component.ts
```diff
@@ -0,0 +1,89 @@
...
+                  const STind = st
+                      .map(function(eachst) {
+                          return eachst.id;
+                      })
+                      .indexOf(this.selectedStationType[sti]);
+                  if (STind < 0) {this.selectedStationType.splice(sti, 1); }
+              }
+          } else { this.selectedStationType = []; }
+      }*/
+    });
```
kjacobsen16: Same as above, is this going to be used in the future?

kjacobsen16 (changes requested):
Left some comments. I mostly just have some cleanup suggestions before merging it in: (1) Mitch pushed the styling changes, so now the components are using scss instead of css, and you can probably remove some of this styling to stay consistent with what Mitch added (he can probably help you with that, he reformatted a lot of it), (2) for the sidebar dropdowns, I recommend using the `ss-multiselect-dropdown` used in the main sidebar component, then you can have the "Check all" and "Uncheck all" options as well, (3) also keep an eye out for cleanup, or make notes of why you're commenting out sections (rather than deleting it) - easiest way is to look through your [changes in the pull request](https://github.com/USGS-WiM/NSS/pull/247/files)

Other than that, it's a great start :)

kjacobsen16 (approved):
Looks and works great! I think when we get further in, we'll probably restructure a bit. For example, the get stations call can be one function, and the agencies and stationtypes will be optionally added to it so we can have stations filtered by both of those (rather than having separate getstationsbytype and getstationsbyagency functions) and we might want to create a separate "gagestats.service.ts" for the gagestats calls, but this is a great start!

Issue #246, opened 2020-07-29
NSS Computation - Issue with commas in regression type description for csv output


kjacobsen16:
If I compute scenarios for Montana, and Download the following section:
![image](https://user-images.githubusercontent.com/35497731/88818763-b02f4e80-d184-11ea-9ffa-df5e9fc20a77.png)

The commas in the description cause issues with the csv columns (exported to xlsx for github, but it downloads as a csv):

[NE_Plains_Region_Bankfull_Region.xlsx](https://github.com/USGS-WiM/NSS/files/4995718/NE_Plains_Region_Bankfull_Region.xlsx)

PR #254, opened 2020-07-29
Issue #246 - Replaces , with ; to fix csv download

ghost:
I replaced the commas with semicolons because they're equally grammatically incorrect. If keeping the commas is important, I'll find a different way!


ghost:
looks like this:
image

kjacobsen16 (approved):
"equally grammatically incorrect" I love it 😆

Issue #252, opened 2020-07-29
Manage Citations Modal - Edits are sticking


kjacobsen16:
I started an edit and then cancelled, but it kept the changed url until I closed out and reopened the modal (should just be `/nss/`, the `/nss-dev/` wasn't actually saved).
![image](https://user-images.githubusercontent.com/35497731/88831094-7b2af800-d194-11ea-99a9-04d244cf7514.png)

kjacobsen16:
Oh, I just noticed, when I reopened the modal, it had the old one and the new one when I re-filtered. Only refreshing the page cleared the non-saved edit.

harper-wavra:
Completed this issue in pull request #258

Issue #197, opened 2020-05-18
Enhancements to Manage Citations Modal


kjacobsen16:
Both of these will be easier once 102 is merged into dev
- [x] Add edit button next to citations and regression regions

- [x] Add button for adding new citation

- [x]  Disable the edit button next to regressions not in their regions

harper-wavra:
For the first point we currently have a button to edit the citation I am just wondering where a good spot would be for the button to edit a regression region? 

Then for the second point I just though a button near the top of the modal for adding a new citation

![image](https://user-images.githubusercontent.com/59063181/87588109-b7485e00-c6a8-11ea-9911-146deb60d3e4.png)

kjacobsen16:
Hm, I forgot about this. Maybe we move the edit/delete citation buttons to the left of the citation row? And put the edit regression region button next to the left of each regression region? Maybe play around with it a bit and see what makes sense .

We could always wait for the edit regression region button in here. It was more of a nice lead in so you could remove the citation from the regression region easily before deleting it, but not necessary.

I think the add citation button sounds good!

harper-wavra:
Right now the only way to add a new citation is inside the creating/editing a regression region modal. If we add a button for just creating a new citation inside the manage citation modal we (I think) won't want to use the creating/editing a regression region modal because the user wont be adding a regression region only a citation. 

I am wondering what the best way is to add a new citation without that modal. Would I need to create a new addcitation component with a modal for just the citation inputs? Or is there a way to have two separate modals inside the managecitations component (one for managing the citations and one for adding a citation)?

kjacobsen16:
So, I actually misread your message at first and thought you were saying you can only create a citation by using the regression region in the url. So I looked that up and that's right, even though that's not what you were asking..

I don't know why it was set up that way, but looks like I'll have to add a separate endpoint allowing for just adding a citation before you can do this issue.

harper-wavra:
I am having issues setting the region correctly in the edit regression region modal. The way we had it working before it just set the region to the selectedRegion which worked because you could only a edit a regression region after you had selected/were viewing the region it belongs too. Now however when you edit a regression region from the manage citation modal you could be in any of the regions so the wrong region populates in the edit regression region modal(if that makes sense.) I was trying to figure out if the region for the regression is in the response from the services but it looks like it is not, so I am unsure how to populate the region dropdown  in the edit regression region modal with the correct region from the regression region and not just the selectedRegion. Does anyone have any ideas?

[...]

PR #258, opened 2020-07-31
Fixes #218 and #252

### LINKED ISSUE #218: Issues with Manage Citation modal when logged in as Manager
When logged in as Manager the Manage Citation Modal has some issues. It seems if you are logged in as a Manager and you subscribe to the service the response doesn't include regressionRegions like it does when you are logged in as an admin.

As Admin:
image

As Manager:
image

harper-wavra:
From #218 disables the edit button next to citations not in their regions, then I moved the last part of this issue to #197.
From #252 solves issue when cancelling edit citation.


kjacobsen16 (approved):
Looks good!

PR #260, opened 2020-08-04
Fixes #72

### LINKED ISSUE #72: Dealing with short name underscores in MathJax
The equation render seems to be slow when rendering the equations, I think they are kinda fun. But any parameter that has an underscore in the short name loses it's underscore and causes the next letter after the underscore to render as a subscript. You may already be aware of it and have no recourse

harper-wavra:


kjacobsen16 (changes requested):
Can you replicate this for the mathjax shown after computing scenarios in the main page? Looks good for edit scenario display!

harper-wavra:
I think I fixed where you were talking about, let me know if there is another place. @kjacobsen16 
image

harper-wavra:
Mathjax loads quick after compute now, however I am not sure creating a new function just to pass back is the best way?

kjacobsen16 (approved):
Looks good

PR #261, opened 2020-08-05
Fixes #232

### LINKED ISSUE #232: Add text input field to filter stations
Another thing we hadn't incorporated when I made my wireframe - we added a "filterText" parameter to the Stations service endpoint. We'll want to add a text input to allow users to filter stations by text.
https://test.streamstats.usgs.gov/docs/gagestatsservices/#/Stations/GET/Stations
image

Anders-Hopkins:
Search box is operational. It searches through gagestat stations by name.

### DIFF
--- a/src/app/gagestats/gagestats.component.html
+++ b/src/app/gagestats/gagestats.component.html
@@ -11,14 +11,7 @@
     </div>
 
     <span MathJax class="hidden" id="mathJaxLoad">`Preloading`</span> 
-    <div class="col-md-6" style="padding: 0;">
-        <div *ngIf="showStationType" class="app-body-section">
-            <label class="big">Station Types:</label>
-            <h5 class="bold">{{selectedStationType}}</h5>  
-        </div>
-    </div>
-
-
+   
    <div class="hidden-print" >
         <table style="border:solid 1px gray" class="custom-table mtop-lg"> 
             <thead>
--- a/src/app/gagestats/gagestats.component.html
+++ b/src/app/gagestats/gagestats.component.html
@@ -13,14 +13,14 @@
     <span MathJax class="hidden" id="mathJaxLoad">`Preloading`</span> 
     <div class="col-md-6" style="padding: 0;">
         <div *ngIf="showStationType" class="app-body-section">
-            <label class="big">Station Type:</label>
-            <h5 class="bold">{{selectedStationType.name}}</h5>  
+            <label class="big">Station Types:</label>
+            <h5 class="bold">{{selectedStationType}}</h5>  
         </div>
     </div>
 
 
-   <div class="hidden-print" *ngIf="showStationType">
-        <table style="border:solid 1px gray" class="table"> 
+   <div class="hidden-print" >
+        <table style="border:solid 1px gray" class="custom-table mtop-lg"> 
             <thead>
                 <tr>
                     <th class="col-xs-1">Type</th>
--- a/src/app/gagestats/gagestats.component.ts
+++ b/src/app/gagestats/gagestats.component.ts
@@ -54,24 +54,17 @@ export class GagestatsComponent implements OnInit {
     this.title = 'Gage Stats';
     this.timestamp = new Date();
     this.showStationType = false;
+    this._nssService.selectedStationType.subscribe((s: Stationtype) => {
+      this.selectedStationType = s;
+    });
     this._nssService.term.subscribe(t => {
       this.term = t;
-      this._nssService.selectedStationType.subscribe((s: Stationtype) => {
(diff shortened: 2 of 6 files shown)


src/app/shared/services/app.service.ts
```diff
@@ -638,6 +650,25 @@ export class NSSService {
...
+        if (term) {
+            url = "?filterText=" + term
+        }
+        if (id.length > 0) {
+           url2 = "?stationTypes=" + id
+       }
+       if (term && id.length > 0) {
+           url = '', url2 = ''
+           url3 = "?filterText=" + term + "&stationTypes=" + id
+       }
```
kjacobsen16: I'm glad you shortened this to one call! you can cut down a few lines here by doing something like 
```
let url = '?';
if (term) { url += 'filterText=' + term; }
if (id.length > 0) { url += 'stationTypes=' + id; }
```
then you can cut out `url2` and `url3`. Also, it's ok to have something like `stationTypes=` (blank) in your http request urls. It'll still work, so you could honestly just try `const url = '?filterText=' + term + '&stationTypes=' + id.join(',')` or something like that 

Also - it's good practice to end each line with a `;`

src/app/shared/services/app.service.ts
```diff
@@ -638,6 +650,25 @@ export class NSSService {
             .get(this.configSettings.gageStatsBaseURL + this.configSettings.stationsURL + "?stationTypes=" + id + "?agencies=" + id2)
     }
```
kjacobsen16: Since you're combining your filters in the `searchStations` function, you can remove the `getStationsByType` and `getStationsByAgency` functions

src/app/gagestats/gagestats.component.html
```diff
@@ -13,14 +13,14 @@ <h4>{{timestamp | date:"MMMM d, y hh:mma"}}</h4>
     <span MathJax class="hidden" id="mathJaxLoad">`Preloading`</span> 
     <div class="col-md-6" style="padding: 0;">
         <div *ngIf="showStationType" class="app-body-section">
-            <label class="big">Station Type:</label>
-            <h5 class="bold">{{selectedStationType.name}}</h5>  
+            <label class="big">Station Types:</label>
+            <h5 class="bold">{{selectedStationType}}</h5>
```
kjacobsen16: It'd be nice to show the name here instead, but that could be a separate issue

src/app/gagestats/gagestats.component.ts
```diff
@@ -53,20 +54,25 @@ export class GagestatsComponent implements OnInit {
...
-      if (s) { this.showStationType = true; }
-      window.scrollTo(0, 0);
+    this._nssService.term.subscribe(t => {
+      this.term = t;
+      this._nssService.selectedStationType.subscribe((s: Stationtype) => {
+        this.selectedStationType = s;
+        if (this.selectedStationType > []) {
+          this.showStationType = true;
+        }
+      });
```
kjacobsen16: Usually I try not to nest subscribers inside each other. I'm guessing you did this because you're calling searchStations here. Instead, when you hit `search` in the sidebar it can actually call the `searchStations` function and create a subject for the selected stations, which you can subscribe to here as well. Hope that makes sense!

Also, the station types get updated in the main gagestats component before the search is triggered, so pushing the search button should also update the stationtype subject.

kjacobsen16 (changes requested):
Works well - I commented mostly on cleanup. I only had one real issue, which is the station types in the main gagestats page (with the station table) update without pushing "Search", but the stations don't get re-queried until you hit search. Honestly, you could just remove that label from the main page

src/app/shared/services/app.service.ts
Anders-Hopkins: Used the `const url` line that you suggested. `term` returns 'undefined' when a station type is not selected, so I added an if/else statement to fix this.

src/app/gagestats/gagestats.component.html
Anders-Hopkins: For some reason I have yet to figure out, selectedStationType is a number rather than an array. Because of this `selectedStationType.name` won't return anything. I removed this pop-up from the main gage stats page. Also, in a later pull request Harper added a function which calls the station type for each selected station and displays this in the station table.

src/app/gagestats/gagestats.component.ts
Anders-Hopkins: I put the `selectedStationType.subscribe` function in with the search button in `gs-sidebar.component.ts` so that the stations are filtered by type when the button is clicked.

kjacobsen16 (approved):

PR #264, opened 2020-08-10
Fixes #256

### LINKED ISSUE #256: Add in pagination of stations
The stations return by default sends through the first 50 stations that match the query (if no filters are given, just the first 50 stations in the DB). The x-wim-messages response header tells you which page was given and how many are available. You can access that header by replicating part of the `outputWimMessages` function. We should:

- [x] show the page number and total available pages at the top of the gagestats page

- [x] allow the user to move to the next page, or type in a page number to go to (I think there are some WIM projects that have similar functionality, @HansVraga any ideas?)

The page number and page count are optional parameters. Default page count is 50, but we can adjust that using the pageCount query parameter. The service documentation shows all optional params: https://test.streamstats.usgs.gov/docs/gagestatsservices/#/Stations/GET/Stations

harper-wavra:
Added in pagination of stations along with a little clean up.

### DIFF
--- a/src/app/gagestats/gagestats.component.html
+++ b/src/app/gagestats/gagestats.component.html
@@ -45,7 +45,7 @@
                     <td>{{s.name}}</td>
                     <td>{{s.location.coordinates[1]}}</td>
                     <td>{{s.location.coordinates[0]}}</td>
-                    <td >{{s.isRegulated}}</td>
+                    <td>{{s.isRegulated}}</td>
                 </tr>
             </tbody>
         </table>
--- a/src/app/gagestats/gagestats.component.html
+++ b/src/app/gagestats/gagestats.component.html
@@ -40,7 +40,7 @@
                     <td>{{s.name}}</td>
                     <td>{{s.location.coordinates[1]}}</td>
                     <td>{{s.location.coordinates[0]}}</td>
-                    <td style="checkbox:">{{s.isRegulated}}</td>
+                    <td>{{s.isRegulated}}</td>
                 </tr>
             </tbody>
         </table>
--- a/src/app/gagestats/gagestats.component.html
+++ b/src/app/gagestats/gagestats.component.html
@@ -12,34 +12,34 @@
    
    <div class="hidden-print" >
         <div class="mtop-lg">
-            <i class="fas fa-angle-double-left" (click)="newPage(1)" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
-            <i class="fas fa-angle-left" (click)="newPage(currentPageNumber-1)" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
-            <input style="width:5%"  [value]="currentPageNumber" (change)="newPage($event)"/>
+            <i title="Click to go to first page" class="icon fas fa-angle-double-left" (click)="newPage(1)" aria-hidden="true"></i>
+            <i title="Click to go backward" class="icon fas fa-angle-left" (click)="newPage(currentPageNumber-1)" aria-hidden="true"></i>
+            <input style="width:5%"  [value]="currentPageNumber" (change)="newPage($event.target.value)"/>
             <label> /{{lastPageNumber}}</label>
-            <i title="Click to go forward" (click)="newPage(currentPageNumber+1)" class="fas fa-angle-right" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
-            <i title="Click to go forward" (click)="newPage(lastPageNumber)" class="fas fa-angle-double-right" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
+            <i title="Click to go forward" (click)="newPage(currentPageNumber+1)" class="icon fas fa-angle-right" aria-hidden="true"></i>
+            <i title="Click to go to last page" (click)="newPage(lastPageNumber)" class="icon fas fa-angle-double-right" aria-hidden="true"></i>
         </div>
         <table style="border:solid 1px gray" class="custom-table"> 
             <thead>
                 <tr>
-                    <th class="col-xs-1">Type</th>
-                    <th class="col-xs-2">Agency</th>
                     <th class="col-xs-1">Code</th>
                     <th class="col-xs-4">Name</th>
                     <th class="col-xs-1">Latitude</th>
                     <th class="col-xs-1">Longitude</th>
+                    <th class="col-xs-1">Type</th>
+                    <th class="col-xs-2">Agency</th>
                     <th class="col-xs-1">Regulated</th>
                 </tr>
             </thead>
             <tbody>
                 <tr *ngFor = "let s of selectedStations">
-                    <td>{{getStationType(s.stationTypeID)}}</td>
-                    <td>{{getAgencyName(s.agencyID)}}</td>
                     <td>{{s.code}}</td>
                     <td>{{s.name}}</td>
                     <td>{{s.location.coordinates[1]}}</td>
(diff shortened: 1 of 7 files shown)


kjacobsen16 (commented):
Anders has a lot of changes in another pull request (https://github.com/USGS-WiM/NSS/pull/261). I started commenting on things and realized he already fixed a lot of it in his branch. Once that gets approved, there are quite a few changes that you'll have to pull in and work through conflicts on. I'll wait until that's done before reviewing this.

kjacobsen16:
@harper-wavra I merged Anders' pull request (#261). There are a lot of changes to the sections you edited for this. You'll have to pull in dev and work through the merge conflicts. If you're unsure of a change, you can look at the pull request, or ask me about it!

src/app/gagestats/gagestats.component.html
```diff
@@ -9,11 +9,17 @@ <h4>{{timestamp | date:"MMMM d, y hh:mma"}}</h4>
         <button *ngIf="loggedInRole" class="button small mright-xs" (click)="bulkUpload()">Bulk Upload</button>
         <button *ngIf="loggedInRole" type="button" class="button small mright-xs" (click)="export()">Export</button>
     </div>
-
-    <span MathJax class="hidden" id="mathJaxLoad">`Preloading`</span> 
    
    <div class="hidden-print" >
-        <table style="border:solid 1px gray" class="custom-table mtop-lg"> 
+        <div class="mtop-lg">
+            <i class="fas fa-angle-double-left" (click)="newPage(1)" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
```
kjacobsen16: I know Mitch doesn't like in-line styling (styling in the html using the style property). Instead, you should add classes to the css and use them here, especially since most of this is repeated.

src/app/gagestats/gagestats.component.html
```diff
@@ -9,11 +9,17 @@ <h4>{{timestamp | date:"MMMM d, y hh:mma"}}</h4>
...
+            <i class="fas fa-angle-left" (click)="newPage(currentPageNumber-1)" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
+            <input style="width:5%"  [value]="currentPageNumber" (change)="newPage($event)"/>
```
kjacobsen16: Here, you should be able to just send `$event.target.value` through. That way you can take out all that if/else stuff and `pageNumber` var from the `newPage` function and just use the event sent through.

src/app/gagestats/gagestats.component.html
```diff
@@ -9,11 +9,17 @@ <h4>{{timestamp | date:"MMMM d, y hh:mma"}}</h4>
...
+            <label> /{{lastPageNumber}}</label>
+            <i title="Click to go forward" (click)="newPage(currentPageNumber+1)" class="fas fa-angle-right" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
+            <i title="Click to go forward" (click)="newPage(lastPageNumber)" class="fas fa-angle-double-right" aria-hidden="true" style="cursor:pointer;padding:5px;"></i>
```
kjacobsen16: If you include the titles here, you should include them on the backward ones as well. And maybe adjust the second one to say `Click to go to last page`

src/app/gagestats/gagestats.component.html
```diff
@@ -27,13 +33,13 @@ <h4>{{timestamp | date:"MMMM d, y hh:mma"}}</h4>
...
                 <tr *ngFor = "let s of selectedStations">
-                    <td>{{s.stationTypeID}}</td>
-                    <td>{{s.agencyID}}</td>
+                    <td>{{getStationType(s.stationTypeID)}}</td>
+                    <td>{{getAgencyName(s.agencyID)}}</td>
                     <td>{{s.code}}</td>
                     <td>{{s.name}}</td>
                     <td>{{s.location.coordinates[1]}}</td>
                     <td>{{s.location.coordinates[0]}}</td>
-                    <td style="checkbox:">{{s.isRegulated}}</td>
```
kjacobsen16, line 36: I'm guessing it wasn't working, but in my original mockup, this was displayed as a checkbox.

src/app/gagestats/gagestats.component.ts
```diff
@@ -46,24 +41,81 @@ export class GagestatsComponent implements OnInit {
...
+    this._nssService.stationTypes.subscribe((stationtypes: Array<Agency>) => {
+      this.stationTypes = stationtypes;
+    });
+    //subscribe to page number related information
+    this._nssService.pageResponse.subscribe((pageText: string) => {
+      let response = pageText;
+      this.lastPageNumber = (response.slice(response.indexOf("of") + "of".length));
+      this.lastPageNumber = (this.lastPageNumber.substr(0, this.lastPageNumber.indexOf('.'))); 
+      this.currentPageNumber = (response.slice(response.indexOf("page") + "page".length));
+      this.currentPageNumber = Number(this.currentPageNumber.substr(0, this.currentPageNumber.indexOf('of')));
```
kjacobsen16: In programming, you can use something called a regular expression (RegEx for short), that allows you to search a string easier than this. Since we just want the numbers from the pageText, you can use the following:
```
var numbers = pageText.match(/[0-9]+/g); // [0-9] means to match any digit, the + means to match where there are multiple digits
      if (numbers != null) {
          this.currentPageNumber = Number(numbers[0]); // first occurrence of a number
          this.lastPageNumber = Number(numbers[1]); // second occurrence of a number
      }
```
To read more on RegEx: https://en.wikipedia.org/wiki/Regular_expression. I use https://www.regextester.com/ to test out my regex before putting it in my code. It also allows you to hover over parts of the regex to see what it means. You could also search by strings like `of` or `page`, but in this context the numbers are the easiest.

src/app/gagestats/gagestats.component.ts
```diff
@@ -46,24 +41,81 @@ export class GagestatsComponent implements OnInit {
...
     this.gagestatsService.addStation();
   }
 
+  public newPage(event){
+    let pageNumber;
+    if(event.target){
+       pageNumber = (event.target.value);
+    }else{
+       pageNumber = event;
+    }
```
kjacobsen16: again, if you just use `$event.target.value` in the html, you don't have to do all this, you can just use `event` throughout, and can remove this pageNumber variable completely.

src/app/gagestats/gagestats.component.ts
```diff
@@ -46,24 +41,81 @@ export class GagestatsComponent implements OnInit {
...
+    if ((pageNumber - 1) * (pageNumber - this.lastPageNumber) <= 0) {
```
kjacobsen16: This is confusing, are you just trying to see if pageNumber is greater than 0 and less than or equal to the last page number?

src/app/gagestats/gagestats.component.ts
```diff
@@ -46,24 +41,81 @@ export class GagestatsComponent implements OnInit {
...
+  public getAgencyName(aID) {
+    let acencyName;
+    if (this.agencies) {
+      this.agencies.forEach(z => {
+          if (aID === z.id) {
+            acencyName = z.name;
+          }
+      });
+    }
+    return acencyName;
```
kjacobsen16: This could all be done on one line (and avoid the forEach loop) by using the `.find()` function (same below in getStationType)

kjacobsen16 (approved):
As usual, just left some cleanup suggestions. Otherwise, it's working well!

Also, if you're going back in, I think in the stations table the stationtype and agency should be after the lat/long, before the Regulated column

Issue #250, opened 2020-07-29
Limit agencies in sidebar drop down


Anders-Hopkins:
Return only the agencies from the available stations to the agency drop-down in the gagestats sidebar.

kjacobsen16:
I don't know that we can do this like we are with NSS. The stations table is huge (36k stations), so what we get from the services is a paginated return (see the options for pagination in the [service documentation](https://test.streamstats.usgs.gov/docs/gagestatsservices/#/Stations/GET/Stations)). So we don't currently have a way of knowing what agencies are used in the available stations. I think just having the entire list of agencies is fine, since it's a small list anyway.

PR #268, opened 2020-08-17
Fixes #248, #250

### LINKED ISSUE #248: Filter gagestats table by agency parameter
Filter the table that is produced from the gagestats sidebar by the agency parameter.

Anders-Hopkins:
This adds agencies to the searchStations() function. #250 can be closed since we do not want to limit the agency options in the dropdown.


kjacobsen16 (approved):
Looks good!

PR #269, opened 2020-08-18
Fixes #267

### LINKED ISSUE #267: Selecting item per page
Have an option for the user to select the number of stations that are shown per page in the gage stats table. Maybe something like: 
 
image

The page count is an optional parameters. Default page count is 50, but we can adjust that using the pageCount query parameter. https://test.streamstats.usgs.gov/docs/gagestatsservices/#/Stations/GET/Stations

harper-wavra:
Adds option to select item per page


kjacobsen16 (approved):
Works great!

PR #273, opened 2020-08-21
Fixes #137

### LINKED ISSUE #137: Create gage page modal
When clicking on a station from the table, a gage page modal should open.

Can add filler text for now until the StreamStats app creates a new gage page

harper-wavra:
Created and populates gage page modal. 
Once approved I will create an issue for editing/adding parameters for the gages.


src/app/gagestats/gagepage/gagepage.component.ts
```diff
@@ -0,0 +1,84 @@
...
+      if (result.show) { 
+          this.code = result.gageCode;
+          this._nssService.getGagePageInfo(this.code);
+        }
+    });
+    this._nssService.GageInfo.subscribe((s: Station) => {
+      this.gage = s;
+      this.getCitations();
+      this.showGagePageForm();
+    });
```
kjacobsen16: Do we need the gage info anywhere else? if not, we can remove the subscriber and just have a function in the nss service that makes the call and returns the gage info and remove the subject/observable. So line 34 would be more like `this.gage = this._nssService.getGagePageInfo(this.code);`

src/app/gagestats/gagepage/gagepage.component.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.modalElement = this.gagePageModal;
+  }
+
+  public getGagePageInfo(){
+    return this._http
+        .get(this.configSettings.gageStatsBaseURL + this.configSettings.stationsURL + '/' + this.code)
+        .subscribe((res: Station) => {
+          this.gage = res;
+        })
+  }
```
kjacobsen16: This looks like duplication of the stuff above with the nssService `getGagePageInfo()` function

src/app/gagestats/gagepage/gagepage.component.ts
```diff
@@ -0,0 +1,84 @@
...
+  
+  public getCitations(){
+    this.gage.citations = [];
+    this.gage.characteristics.forEach(c => {
+      if (c.citationID) {
+         if (!this.checkForDupCitations(c.citationID)) {
+          this.gage.citations.push(c.citation);
+        }
+      }
```
kjacobsen16: No need for a separate function (or two if statements), since the function is basically one line. Instead you can say:
`if (c.citationID && !this.gage.citations.some(cit => cit.id === c.citationID))`
Same below for the statistics

Nice use of the some function, btw!

src/app/gagestats/gagepage/gagepage.component.html
```diff
@@ -0,0 +1,138 @@
...
+					<div>
+						<table style="border:solid 1px gray" class="custom-table"> 
+							<thead>
+								<tr>
+									<th>Statistic ID</th>
+									<th>Value</th>
+									<th>Units</th>
+									<th>Citation ID</th>
+									<!-- <th>Preffered?</th>
+									<th>Years of Record</th> -->
```
kjacobsen16: These are both sent back in the getstation endpoint (although years of record is not attached when null)

kjacobsen16 (changes requested):
Looks like a great start! A few small things: (1) Can you change "Characteristic ID" and "Statistic ID" to "Characteristic" and "Statistic"? (2) When you mouse over the table row, we should change the pointer and color of the table row so you know it's clickable (Mitch could probably do this pretty quickly).  (3) Not sure if it matters, but there are a few things missing from your interfaces, like the characteristic -> variabletype and statistic -> regressiontype, and unit types for those as well. If you run the get station query in insomnia, it's easier to view/collapse the properties and check everything. Nice job :)

kjacobsen16 (approved):
Looks great!

Issue #253, opened 2020-07-29
Styling fixes


kjacobsen16:
- [x]  Variables dropdown errors feel a little smushed in add scenario modal (to replicate - go to add scenario modal and hit the "Add Parameter" button
![image](https://user-images.githubusercontent.com/35497731/88830067-102cf180-d193-11ea-9421-3565b7096c7e.png)

- [x] Manage citations modal column headers are off 
![image](https://user-images.githubusercontent.com/35497731/88830906-3a32e380-d194-11ea-984f-1c46955fc7f2.png)

kjacobsen16:
@harper-wavra if you notice any small styling issues, you can add them here

harper-wavra:
- [x] When not logged in the icons on the sidebar are spaced differently
[image](https://user-images.githubusercontent.com/59063181/88832156-048efa00-d196-11ea-8d8e-fa18f38650ef.png)
- [x] In settings the New Regression Type and New Station Type modals are missing the new button classes
[New Regression Type](https://user-images.githubusercontent.com/59063181/88832865-0c02d300-d197-11ea-920e-4942b0c9e738.png)
[New Station Type](https://user-images.githubusercontent.com/59063181/88832909-1b821c00-d197-11ea-8306-6c00d568e9c1.png)

kjacobsen16:
New issue: the equations section (shown after computing scenarios on the main page) runs together. Before, each equation was on a separate line
![image](https://user-images.githubusercontent.com/35497731/89039470-c1ed2f00-d307-11ea-85b5-e87894c00380.png)

PR #281, opened 2020-09-02
Issue #253 and other layout and styling fixes.

mitchas:


harper-wavra:
Looks good! The only thing I noticed was the manage citations modal is still a little wonky when you filter it.
image

harper-wavra:
It seems to look just a little bit off when the scroll bar is available, it seems to push the columns to the left a little bit, and when the citation is less than a full line of text (which is only our test citations so I don't think really this matters.) Might have to see what @kjacobsen16 thinks, but I think it looks good.

[scroll bar](https://user-images.githubusercontent.com/59063181/92256688-a55d8d00-ee99-11ea-93cb-9c0df350a3b7.png)
[less than full line of text](https://user-images.githubusercontent.com/59063181/92256721-b1494f00-ee99-11ea-817c-f8545d8e1548.png)

kjacobsen16:
@mitchas one more thing I noticed - our form validation styling seems to have been removed. To replicate: log in, go to Settings, click on Agencies, "Add New Agency", then start typing in the Name field and remove the text. You should see this text below:
image

Originally, the input box would have a red outline around it, and the text underneath would be red. I believe we were using the `form-invalid`, etc. classes. They should all be in the html still.

kjacobsen16 (commented):
Looks good! I made a couple quick edits, but there are two more items that should be fixed:
1. The main NSS page is only printing one page. To replicate: on main page, select any region, put values in for each explanatory variable, hit "compute", then try to print the page.
2. The map is overlapping a button in the Add Regression Region modal. To replicate: log in, in the main page hit the plus icon next to "limit by hydrologic region", click the button at the bottom that says "Upload Polygon".
image

If you want, we can make those a separate issue.

harper-wavra (approved):
Looks good, going to merge and create a new issue with the two things KJ noticed.

PR #283, opened 2020-09-03
Fixes #270

### LINKED ISSUE #270: Issues when adding errors
Issues relate to edit scenario page, edit regression, under the Errors column  

- [x] “Add error” option should be edited to look like it is a hyperlink or button

- [x] When it is selected, the previously entered error is removed.

harper-wavra:
“Add error” option made to look like button

 Resolved  previously entered error being removed


kjacobsen16 (approved):
Looks good!

PR #289, opened 2020-09-10
Fixes 277

harper-wavra:
When the statistic group has been selected in the Clone/New Scenario modal the regression variables are filtered to that group. Added an option to remove the filter. In settings also added an option to filter.


src/app/settings/categories/regressiontypes/regressiontypes.component.ts
```diff
@@ -45,6 +46,8 @@ export class RegressionTypesComponent implements OnInit, OnDestroy {
     public tempData;
     public isEditing = false;
     public modalRef;
+    public statisticGroups: Array<Statisticgroup>;
+    public selectedSatistic;
```
kjacobsen16: Just to eliminate confusion here, I think this should be `selectedStatisticGroup`, because `statistic` generally refers to the regression type, or regression variable (or in the case of gagestats, the actual calculated statistic).

src/app/settings/categories/regressiontypes/regressiontypes.component.ts
```diff
@@ -88,6 +97,18 @@ export class RegressionTypesComponent implements OnInit, OnDestroy {
                 });
         }
     }
+    public onSatSelect(e){
```
kjacobsen16: Maybe change this to `onStatGroupSelect` or something like that

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -349,6 +352,13 @@ export class AddScenarioModal implements OnInit, OnDestroy {
         this.addPredInt = false
     }
 
+    public onSatSelect(e){
```
kjacobsen16: Again, should be `onStatGroupSelect`

kjacobsen16 (changes requested):
This works great, I think Pete will really like it! A couple ideas:
- [x] We should add a tooltip (like the one below on the main page) that explains what the auto filter does, something like "Filters the regression variables by the selected statistic group. Removing the filter shows all available regression variables."
image
- [x] Let's show all statistics if no statistic group has been selected yet.
- [x] Can you combine the region and stat group filters in the reg types settings page? The user might not notice that the other filter has changed to "None" and may expect both filters to be applied

kjacobsen16 (approved):
Works great! One note - with the filters on the regression type page, instead of comparing the lists, you could actually just send both in the url for the services call, e.g. https://test.streamstats.usgs.gov/nssservices/regressiontypes?regions=2&statisticgroups=2. If one is not filtered, you can just send nothing on the other side of the equal sign:
https://test.streamstats.usgs.gov/nssservices/regressiontypes?regions=&statisticgroups=2
https://test.streamstats.usgs.gov/nssservices/regressiontypes?regions=5&statisticgroups=

PR #295, opened 2020-09-16
Fixes #287

### LINKED ISSUE #287: Display region in gagestats table
Stations now have a "regionID", so we can use that to display the name of the region.

There is also now a `regions` endpoint where you can access all the regions (uses the same table as NSS)

harper-wavra:
added region to gagestats table

### DIFF
--- a/src/app/gagestats/gagestats.component.html
+++ b/src/app/gagestats/gagestats.component.html
@@ -28,6 +28,7 @@
                 <tr>
                     <th class="col-xs-1">Code</th>
                     <th class="col-xs-4">Name</th>
+                    <th class="col-xs-4">Region</th>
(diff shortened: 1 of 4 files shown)


harper-wavra:
I used the NSS Services (https://test.streamstats.usgs.gov/nssservicesv2/regions) since it was already there but maybe I should use the GageStats Services (https://test.streamstats.usgs.gov/gagestatsservices/regions)?

kjacobsen16:
They point to the same table in the database, but for consistency I think it would be best to use the gagestatsservices in the gagestats portion.

Also - sorry I haven't had time to review these, I'm hoping to have more time for NSS next week.

kjacobsen16 (approved):

PR #298, opened 2020-09-21
Fixes #275

### LINKED ISSUE #275: Make gagepage editable
Gagepage should be editable. There should be buttons to add a new char/stat as well as buttons (maybe like when you edit a regression) to edit a char/stat.
 
image

Anders-Hopkins:
Makes the characteristics and statistics tables within the gagepage modal editable.


kjacobsen16 (changes requested):
Great start! There are some improvements we could make, I think some more info would help :slightly_smiling_face: 

So, to explain the properties in the station return further - the nested objects, like the citation, unittype, variabletype, etc. are purely informational, it just allows us to see that info without having to make a separate call to get it. The services get those objects by looking at the ID properties, such as the citationID, unitTypeID, variableTypeID, etc. So when you create a new characteristic or statistic you don't need to attach the full objects, just the ID. In the case of a new characteristic, you'd just need the value, comments, variableTypeID, unitTypeId and citationID (and of course stationID, which we already have). In the interface for the gagechar object, you can make those optional by adding a questionmark, e.g. unitType?: Unittype.

The issue with the IDs are the user doesn't actually know what IDs go to what items, so we'll need a dropdown to show them the name of the object and allow them to select it, but then use it to send through the ID. You've almost got that set up for the unit type! The only change that needs to happen there is changing the ngModel and ngValue properties so that it updates the ID instead of the attached object, and then we can remove the input for the unit type ID.

Another improvement I thought of - instead of adding the new char/stat to the list, we can actually add a new table row with the new char/stat after the <tr> with the ngFor in it, so after line 153 add something like <tr *ngIf="newChar">. Then we'd read and set the properties by using the newChar variable, instead of the c variable. That, along with the change to make the nested objects optional, means all you have to do in the addPhysicalCharacteristic() function is just create the newChar object, something like:
```
this.newChar = {
      stationID: this.gage.id,
      value: null,
      comments: null,
      variableTypeID: null,
      unitTypeID: null,
      citationID: null
    }
```
The same would apply for the new statistics, and it'd be very similar for editing chars/stats :slightly_smiling_face:. That's a lot of info and maybe not super clear, so feel free to ask questions, or I can jump on a call with you/explain more of this to you.

Anders-Hopkins:
This makes sense to me for the most part. The problem I discovered with not including a nested objects for "units" in the newChar variable is that the units dropdown menu would not work once the newChar is added to the table. Is there a work around for this? The same would apply once we add dropdown menus for other variables.

kjacobsen16:
I think if you change the ngModel and ngValue objects it should work, so something like:
```
<select [compareWith]="compareObjects" name="unitType" [(ngModel)]="newChar.unitTypeID">
	<option *ngFor="let unit of units" [ngValue]="unit.id">{{unit.unit}}</option>
</select>
```

kjacobsen16:
So the ngModel property is telling it to bind the selected item to the newChar's unittypeid, the ngValue is saying to set it to the id of the selected unit from the dropdown

Anders-Hopkins:
I've made the changes that you suggested. I removed the unitTypeID from both the characteristic and statistic tables as well as the regressionTypeID from the statistics table. The drop down menus for units will set the unitTypeID in the put/post request, just as the regressionType (seen in the statistics table as the 'Statistic' column) sets the regressionTypeID for these requests. 

I was not able to include a dropdown menu for the statisticGroupType because this object in not included in the get station response. 
gagepagemodal

kjacobsen16 (approved):
Works great! I think it still could use some styling help. I can't see some of my inputs when I'm adding or editing a statistic.

Also, Mitch usually recommends using classes rather than inline styling (using the `style` tag in the html) because it's difficult to pinpoint issues caused by inline styling. He has some good preset width classes in the `styles/data.scss` file that use percentage, or you can create your own classes.

kjacobsen16 (approved):
I added an additional comment [here](https://github.com/USGS-WiM/NSS/issues/304) to fix the styling issues, so going to go ahead and merge. Nice job!

Issue #265, opened 2020-08-12
Appendix is duplicating


kjacobsen16:
Not sure when this started happening, but if you select "Test" as the region, then calculate scenarios, you'll see the appendix multiple times (seems like one for each statistic group)

PR #303, opened 2020-09-23
Issue #265 - Prevents appendix from duplicating

ghost:
All it took was moving one div(!!)
The issue said to use 'Test' to test it, but I think that region has since been edited, so it's doesn't work for testing this anymore. All of the regions duplicated the appendix if you selected at least 2 statistic groups, so if you test it, just use any of those regions (like Florida or Connecticut).


kjacobsen16 (approved):
Looks good - thanks for the info on how to test it, it helped speed things up!

PR #309, opened 2020-09-28
Update "Finish Edits" button 

Anders-Hopkins:
When "Finish Edits" is clicked, it asks to delete unsave new chararcteristics or new stats. Also cancels editability of chars or stats being edited.


kjacobsen16 (approved):
👍

PR #310, opened 2020-09-28
Fixes #128

### LINKED ISSUE #128: Add Default Unit to New Variable Types
I need to check and see if this is set up in the services yet, but we should allow the user to select a unit when creating a variable type.

Also, that default unit should be automatically selected out of the dropdown when a variable is selected for adding a scenario (again, need to check services)

- [x] allow the user to select a default unit when creating a variable type.
- [x] default unit should be automatically selected out of the dropdown when a variable is selected for adding a scenario.

harper-wavra:
allows the user to select a default unit when creating a variable type.
default unit is automatically selected out of the dropdown when a variable is selected for adding a scenario.


kjacobsen16 (changes requested):
Looking good! A couple things I noticed:
1. The add scenario unit selection isn't quite working right, I think it's because the html is sending the code but you're comparing it to the variable's name.
2. I noticed when I added a new variable then went to add a new scenario it didn't have the new variable in the dropdown yet. Might be an easy fix, but feel free to add a new issue for it if not.

Also, can you use the defaultunitid to display the default unit in the variable type table in the settings page (if it exists) and make it editable? My bad - I should've added it to the issue!

kjacobsen16 (approved):
Works great!

PR #311, opened 2020-09-28
Fixes #285

### LINKED ISSUE #285: Add loader to "Edit Regression Region" Modal
Loading is slow after opening the edit regression region modal a couple times, especially for reg regions with polygons

The loader is already used a few places in the app, should be pretty easy to incorporate

harper-wavra:
Loader is working again (small fix).


kjacobsen16 (approved):
It's always the simplest stuff 😅 nice job

Issue #304, opened 2020-09-23
Styling fixes


harper-wavra:
- [x] The main NSS page is only printing one page. To replicate: on main page, select any region, put values in for each explanatory variable, hit "compute", then try to print the page.
- [x] The map is overlapping a button in the Add Regression Region modal. To replicate: log in, in the main page hit the plus icon next to "limit by hydrologic region", click the button at the bottom that says "Upload Polygon".
![image](https://user-images.githubusercontent.com/59063181/94050316-b49a7100-fd9b-11ea-8260-fd13b20bc667.png)

- [x] Gage page edit issues (see comment by KJ below)

kjacobsen16:
New item - we added some functionality to the gage page modal, steps to replicate: (1) log in, (2) type "testing" into search by keyword and hit "search" at the bottom of the sidebar, (3) click on the test station in the station table, (4) click "Edit Gage" at the top of the gage page modal, (5) try "add statistic" and "edit statistic" (pencil icon next to stat) and plug in some values

When we have select dropdowns, it affects the width of the other table cells. I tried to work with the word wrap and white space properties but no luck so far.

mitchas:
Everything here should be fixed now - as well as a few other small things I noticed.

As far as I can tell the printing is working as it should now - I added a new print.scss file that can be used to adjust any styles for printing

PR #313, opened 2020-09-30
Ms 304

mitchas:
Fixes issues in #304 

Added a custom class to make bigger (wider) modals. 

If any other modals sizes need to be increased, you can add the custom class modal-xl with the windowClass parameter:
```
    this._modalService.open(this.modalElement, { backdrop: 'static', keyboard: false, size: 'lg', windowClass: 'modal-xl' });
```


kjacobsen16 (approved):
Looks good, thanks Mitch!

PR #330, opened 2020-10-13
Fixes #316

### LINKED ISSUE #316: Station Edits
- [x] Allow user to edit the general station details such as name, code, station type, etc.

- [x] Allow user to delete gage

Since we already have the "Edit Gage" button in the gage page modal, that might be the place for it.

harper-wavra:
Add option to delete gage
Add option to edit gage information
Minor text changes to page info
Requerys filters when deleting gage or editing gage information


kjacobsen16 (commented):
Looks good - one thought I had was that the filter items could probably be combined into one object with a property for each of the filter items, rather than having an individual variable and observable/subject for each item.

Another thought - maybe we can play with filtering automatically on selection and see if that's fast enough (rather than having to hit "Search", since on the NSS side it filters on select). If that's easy enough to try we can see if it works - if not, no big deal.

harper-wavra:
Put all the filter options into a object. There is probably a better way to do the `on_____Select` functions instead of having separate ones for each filter but I wasn't sure what that would be.

kjacobsen16:
Hm, instead of having the `On...select`, you could have the html update the params property for that filter item, then have the onSearch function update the variables in the service?

Edit: then you can also call the `onSearch` function in the html. There also might be ways to use angular models to trigger it, but I don't think we need to spend too much time on this 🤷

kjacobsen16:
Seems like the queries are pretty fast though, so I think not having to hit the "Search" button is an improvement :) thanks!

harper-wavra:
Yeah that worked well, thanks!

kjacobsen16 (approved):
Looks good!

PR #332, opened 2020-10-14
Fixes #312

### LINKED ISSUE #312: "Refresh" when creating new items in settings
When you add a new variable/error/test ect in the settings then go to add a new scenario it doesn't have the new variable/error/test ect in the dropdowns yet.

harper-wavra:
Created function for the getEntities in order to be able to call it again before the modal shows.
Also changed UnitsSystems to UnitSystems.


kjacobsen16 (approved):
👍

PR #340, opened 2020-10-20
Fixes #328

### LINKED ISSUE #328: Add Station Modal Changes
When gage is added successfully we should maybe:

- [x] clear the modal inputs

- [x] close the modal?

- [x] requery gages with current filters

☝️  I think this is how the add scenario modal does it, we should be consistent with that

harper-wavra:
clears the modal inputs, closes the modal, requerys gages with current filters

Also fix from issue #316, forgot to call `this._nssService.setSelectedFilterParams(this.params);`


kjacobsen16 (approved):
Works well!

PR #345, opened 2020-10-20
Fixes #317

### LINKED ISSUE #317: Rename "Regions" in the client to "Study Areas"
Each reference to "Regions" should actually be "Study Areas". This is separate from the Regression Regions, those should be stay the same.

harper-wavra:
I think I got them all


kjacobsen16 (approved):
Looks good

PR #346, opened 2020-10-21
Fixes #342 - Align edit icons in Citation Manager

### LINKED ISSUE #342: Align edit icon in Manage Citations
For some of the citations, the horizonal alignment of the edit icon gets off enough that it's hard to tell which regression region it corresponds with.
image

Steps to reproduce:
1. Login 
2. NSS
3. Select Rainy River Basin
4. Click 'Manage Citations'
5. Scroll down and notice the alignment is off (most obvious case is at the end of the 'Thomas' citation)

ghost:
The solution I found partitions the edit icons and associated regression regions into semi-separated cells. It has a little bit of a different look before, but the icons and regression regions are now always aligned:
image

The alignment for all the vertical columns was also a little off before - not sure why the existing percentages didn't work before, but I hardcoded widths - I am hoping but not certain this fix will translate over to other monitor sizes.

### DIFF
--- a/src/app/shared/components/managecitations/managecitations.component.html
+++ b/src/app/shared/components/managecitations/managecitations.component.html
@@ -1,8 +1,8 @@
 <ng-template #manageCitations let-c="close" let-d="dismiss">
 	<div>
 		<div class="modal-header">
-			<div class="title">Manage Citations</div>  
-			<button class="icon-button" (click)="d('closed')"><i class="far fa-times"></i></button>                
+			<div class="title">Manage Citations</div>
+			<button class="icon-button" (click)="d('closed')"><i class="far fa-times"></i></button>
 		</div>
 
 		<div class="modal-body modal-body-fixed side-padding">
@@ -10,7 +10,7 @@
 			<div>
 				<input type="text" #input placeholder="Search" (keyup)="filter(input.value)">
 			</div>
-			<div  [class.hidden]="showNewCitation" class = "mtop-sm">
+			<div [class.hidden]="showNewCitation" class="mtop-sm">
 				<button type="button" (click)="showNewCitation = true" class="button blue">Create New Citation</button>
 			</div>
 			<form *ngIf="showNewCitation" [formGroup]="newCitForm" class="newCitForm">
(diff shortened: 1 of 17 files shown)


kjacobsen16 (changes requested):
It looks a lot better! I think we can clean it up a bit though. Some suggestions:
1. Generally, it's better to style using classes or ids, rather than using the `style= xx` in the html (called inline styling), because it's more difficult to find the sources of styling that way.
2. Instead of having two separate sections for the `isLast` property in the html, you can have one block of code but change the styling with conditional classes. Similar to how we use `[class.hidden]` to say "give this the css `.hidden` class if the statement is true", we can do that for other classes. So you can move the styling to the scss file, make a class in the scss for `.isLast` and use it in the html like `[class.isLast]="isLast"`, etc. Hope that all makes sense!

ghost:
@kjacobsen16 , thanks for the comments - it made the code look a lot less jumbled! I think it's ready now

ghost:
(I accidently pressed the close button when I commented - reopened now)

kjacobsen16:
Wow, you're speedy!

src/app/shared/components/managecitations/managecitations.component.scss
```diff
@@ -2,10 +2,37 @@
...
     margin: 0px 5px 5px 0px;
 }
 .modal-body{
     overflow-y: auto;
 }
+
+.regRegion {
+    width: 228px; 
+    border-top-style:none; 
+    border-left-style:none;
```
kjacobsen16, line 15: These could be moved to a common css styling since they're used in all of them - this can be done by setting the class for the div that all the regRegion/editRegRegion, etc. are in. Then, say you title it `class="regRegionDiv"` or something like that, you can then set the styling you want all the tds inside it to have by using:
```
.regRegionDiv td {
    border-top-style: none;
    border-left-style: none;
}
```

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -19,99 +19,119 @@
...
 									<i class="far fa-plus"></i>
 								</button>
 							</div>
 						</td>
-						<!-- Col 3, Associated Regression Region and Buttons -->
-						<td class="width-35">
-							<div *ngFor="let e of c.regressionRegions; let lst = last;">{{e.name}}<span *ngIf = "!lst"><hr></span></div>
+						<!-- Col 3, Associated Regression Region -->
+						<td>
+						<div *ngFor="let e of c.regressionRegions; last as isLast">
```
kjacobsen16: See comment below in scss, the `class="regRegionDev"` would go here

kjacobsen16 (commented):
Nice, that's a lot better! There's still some inline styling in the html (I'm viewing the changes by going to https://github.com/USGS-WiM/NSS/pull/346/files ). Also left a couple comments for some helpful tips on styling.

Functionality is good, so I'm approving - these changes are optional

kjacobsen16 (approved):

src/app/shared/components/managecitations/managecitations.component.scss
ghost: Oh, that's a good idea! I was thinking each class was slightly different, but it's true most of them are nearly the same with one attribute added/removed.

ghost:
@kjacobsen16  - I'll make the rest of the html styles in the scss, so you can hold off on merging for a few mins!

ghost:
@kjacobsen16 
Alrighty, I went through and got rid of all the styles= in the html instead of just the ones in the last two columns! I ended up deciding not to do the extra class in the last div because on further look there was only one style that overlapped with all classes and even when I created on just for that, it seemed to get overwritten when I called the other classes after. I also tacked on one other small change: there's now a '+' next to 'Create New Citation' to make it consistent with all the other 'Add ___' buttons in the application.

kjacobsen16:
Thanks Kathy - going ahead and merging now!

PR #349, opened 2020-10-21
Fixes #308 - Makes filters searchable

### LINKED ISSUE #308: Enhancements to Gagestats Sidebar
There can be a lot of options for the sidebar in Gagestats, in order to enhance the experience one or both of the following should be added:

- make the dropdowns searchable
- use the previous selections to limit the options in the other dropdowns (like NSS does), may require additional service work

ghost:
The second part of Issue #308 (make limits filter options based on other selected filters) is now Issue #180 in NSS Services.


harper-wavra (approved):
👍

PR #351, opened 2020-10-23
Fixes #131

### LINKED ISSUE #131: Duplicate/multiple requests?
It seems that there are a LOT of requests going through when a region is selected, many of them duplicates. This might be an issue with how the selected region is being triggered.

There are quite a few requests on initial selection, but even more after a new state is selected. To replicate:
1. Select a region in the sidebar
2. Open dev tools and clear the console
3. Select a new region
Image below is just from me selecting Rhode Island (a small state).
image

harper-wavra:
Originally the initial region selection had anywhere from 15-20 XHR requests and all selection after that had anywhere from 40-90.

On initial region selection I was able to lower to 7 XHR requests and all following region selections I was able to lower to 25 XHR requests.

### DIFF
--- a/src/app/mainview/mainview.component.ts
+++ b/src/app/mainview/mainview.component.ts
@@ -1494,6 +1494,7 @@ export class MainviewComponent implements OnInit {
     }
 
     public getCitations() {
+        console.log('mainview')
         this._settingsService.getEntities(this.configSettings.citationURL)
             .subscribe(res => {
                 this.citations = res;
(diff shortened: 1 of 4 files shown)


harper-wavra:
I think this is ready for review now.

kjacobsen16 (commented):
Looks pretty good - I am noticing that the manage citations modal isn't getting updated when a citation is added/edited/removed from a regression region, etc. though.

kjacobsen16 (approved):
Looks good, I'm merging this because this can wait, but I think we can still bring the count down. It will need some reworking in the app.service.ts where it sets the selected region and gets all the filter data. As you can see below, it's getting the filter data when a region is selected, but then refiltering each one with empty query parameters
image

Issue #307, opened 2020-09-24
Add Manage Citation Modal to Gagepage


Anders-Hopkins:
Use the Manage Citation Modal to select citations for the characteristics and statistics within the gagepage modal. A button should be added to open the manage citation modal when add/editing chars or stats. Then the citation selected in the modal should set the citationID of the char or stat.

PR #355, opened 2020-10-26
Fixes #354 - Prevents user from editing multiple stats at once

### LINKED ISSUE #354: Gage Page Edit Changes
When you're editing characteristics or statistics, you can click "edit" on multiple lines, but when you save one, it cancels all other edits.

I can see a user coming in and opening multiple characteristics to edit them and not realize that only one will save, meaning lost work for the user.

We need to either only allow the user to edit one item at a time, or allow for a "save all changes" button.

ghost:
When a characteristic is being edited, user must save/delete/finish editing that characteristic before editing another one.

### DIFF
--- a/src/app/gagestats/gagepage/gagepage.component.html
+++ b/src/app/gagestats/gagepage/gagepage.component.html
@@ -126,7 +126,7 @@
 						<tr *ngIf="newChar">
 							<td>	
 								<div class="table-icon-buttons">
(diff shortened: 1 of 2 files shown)


ghost:
@Anders-Hopkins already made it so only one gage is edited at a time in #307 (locally) and it's more efficient, so I'm closing this!

PR #357, opened 2020-10-27
Fixes #348

### LINKED ISSUE #348: Changes to edit scenario
I think the way we edit regression equations should be adapted to make the full scenario editable in one go.

Since we're already cloning the scenarios, I think we can do a similar thing where we send through the scenario to edit to the add scenario modal and instead of cloning, we let them edit and save when done.

harper-wavra:
When editing regression equations the full scenario is now editable in one go.

I tested my changes in the Test region and it seemed to be working how I thought it should without giving me any errors. However, there may be changes/comments on this if I wasn't thinking about it 100% correctly.


harper-wavra:
If any of the key defining features are changed the scenario is cloned instead of edited. When any of those three core dropdowns are changed, I alerted the user with a toast message. There might be a better way then the toast message because if the user changes a dropdown then changes it back the message is similar and they might miss it. 

I tested my changes in the Test region and it seemed to be working how I thought it should without giving me any errors. However, I am expecting changes/comments on this if I wasn't thinking about it 100% correctly.

src/app/mainview/mainview.component.ts
```diff
@@ -1222,6 +1222,13 @@ export class MainviewComponent implements OnInit {
     }
 
     /////////////////////// Edit Scenarios Section ///////////////////////////
+    public editScenarioRowClicked(statisticGroupID, r, rr) {
+        let info = 'edit';
+        this.cloneScen = { r, rr, statisticGroupID, info };
+        this.newCloneScenario(this.cloneScen);
+        this.showCloneScenarioModal();
```
kjacobsen16: A couple things to clean up here:
1. You can just make lines 1225-1226 into one line: `this.cloneScen = { r, rr, statisticGroupID, 'edit' }`, or you can just combine this whole function with the `cloneRowClicked` function and send the `info` variable in the function call.
2. `newCloneScenario(...)` is a one-line function and it's only called twice (once if you combine this with cloneRowClicked), so you can just pull that line out and remove the function

src/app/shared/components/addscenario/addscenario.component.html
```diff
@@ -299,7 +300,9 @@
 			<button type="button" (click)="clearScenario()" class="button red mright-xs">Clear All</button>
 			<button type="button" (click)="c('closed')" class="button black">Cancel</button>
 			<div class="flex-spacer"></div>
-			<button type="button" (click)="createNewScenario()" [disabled]="!newScenForm.valid" class="button blue">Create</button>
+			<button type="button" *ngIf="edit==false" (click)="createNewScenario()" [disabled]="!newScenForm.valid" class="button blue">Create</button>
+			<button type="button" *ngIf="edit==true" (click)="submitScenario()" [disabled]="!newScenForm.valid" class="button blue">Save</button>
```
kjacobsen16, line 304: You could probably just combine these by using one function that has an if/else statement to see if edit = true

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -58,6 +59,11 @@ export class AddScenarioModal implements OnInit, OnDestroy {
     public tempSelectedStatisticGrp: Array<Statisticgroup>;
     public filteredRegressionTypes;
     public filtered = true;
+    public scen;
+    public orgStatGrp;
+    public orgRegRegion;
+    public orgRegVariable;
```
kjacobsen16: I think just having an `originalScenario` variable would be a lot cleaner.

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -182,16 +188,33 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
+            else if (this.cloneParameters.info == "edit"){
+                this.clearScenario();
+                this.clone = false;
+                this.edit = true;
+                this.filtered = false;
+                this.editMode = true;
+                this.orgStatGrp = this.cloneParameters.statisticGroupID;
+                this.orgRegRegion = this.cloneParameters.rr.id;
+                this.orgRegVariable = this.cloneParameters.r.id;
+                this.fillModal();
```
kjacobsen16: Some items are repeated between between the two if/else if statements (`thisClearScenario()`, `this.filtered = false`, etc.)

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -58,6 +59,11 @@ export class AddScenarioModal implements OnInit, OnDestroy {
     public tempSelectedStatisticGrp: Array<Statisticgroup>;
     public filteredRegressionTypes;
     public filtered = true;
+    public scen;
+    public orgStatGrp;
+    public orgRegRegion;
+    public orgRegVariable;
+    public editMode: boolean;
```
kjacobsen16, line 64: Is this different than `edit`?

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -287,10 +306,41 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
+        if (this.editMode == true){
+            this.scen = JSON.parse(JSON.stringify(this.newScenForm.value));
+            //change back to edit if user reselects original core dropdowns
+            if ((parameter == "statGroup" && this.orgStatGrp == this.scen.statisticGroupID) ||
+            (parameter == "regRegion" && this.orgRegRegion == this.scen.regressionRegions.ID) || 
+            (parameter == "regVariable" && this.orgRegVariable == this.scen.regressionRegions.regressions.ID)){
+                this.clone = false;
+                this.edit = true;
+                this._toasterService.clear();
+                this._toasterService.pop('info', 'Info', 'Scenario will be Edited Instead of Cloned');
```
kjacobsen16: Inconsistent capitalization here

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -377,17 +427,47 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
+            .subscribe((response) => {
+                if (this.originalRegion == this.selectedRegion) {
+                    this._nssService.selectedStatGroups = this.tempSelectedStatisticGrp;
+                    this._nssService.setSelectedRegRegions(this.tempSelectedRegressionRegion);
+                    this._nssService.selectedRegressionTypes = this.tempSelectedRegType;
+                } else {
+                    this._nssService.setSelectedRegion(this.selectedRegion);
+                    this._nssService.selectedStatGroups = [];
+                    this._nssService.setSelectedRegRegions([]);
+                    this._nssService.selectedRegressionTypes = [];
```
kjacobsen16: This could be made into a separate function, because this is pretty much exactly what is called after the postentity in `createNewScenario`. Generally the way I treat functions is: if you're calling more than a couple lines more than once, it's easier to create a function. If you're calling just one or two lines, usually it's a waste of a function.

kjacobsen16 (changes requested):
Overall works great - I think this is a big improvement!

This is probably an edge case but I did some playing around with the selects and if I change two of the core identifying properties to new items and then change only one of them back, it says it'll be "edited, not cloned", when at that point since there's still one different item it should be cloned.

The rest of my comments are just basic cleanup, they're not required.

src/app/shared/components/addscenario/addscenario.component.ts
harper-wavra: Do you mean the info's or the 'Scenario will be Edited Instead of Cloned'?

src/app/shared/components/addscenario/addscenario.component.ts
harper-wavra: It is different, yes.

src/app/shared/components/addscenario/addscenario.component.ts
kjacobsen16: I was referring to "Scenario will be Edited Instead of Cloned", as well as the "...Cloned Instead of Edited"

kjacobsen16 (approved):
Works great!

PR #359, opened 2020-10-29
Fixes #307 & #354

### LINKED ISSUE #307: Add Manage Citation Modal to Gagepage

Anders-Hopkins:
Added the Citation Modal to the GagePage so that it can be used to select the citations for new chars/stats and for editing chars/stats. Also made it so that only a single newchar/newstat/char/stat/GageInfo can be edited at a time. Made edits to the Citation Modal when it is opened in the GagePage: added an ID column and removed the last two columns.

### DIFF
--- a/src/app/shared/components/managecitations/managecitations.component.html
+++ b/src/app/shared/components/managecitations/managecitations.component.html
@@ -45,8 +45,8 @@
 					<tr>
 						<th class="width-60">Citation</th>
 						<th class="width-5"></th>
-						<th class="width-35">Associated Regression Regions</th>
-						<th class="width-5 empty-cell">
+						<th class="width-35" *ngIf="!inGagePage">Associated Regression Regions</th>
+						<th class="width-5 empty-cell" *ngIf="!inGagePage">
 							<!-- Filler button - not used for anything -->
 							<div class="cell-buttons"><button class="icon-button" aria-hidden="true"><i class="far fa-circle"></i></button></div>
 						</th>
@@ -55,7 +55,7 @@
 				<tbody>
 					<tr *ngFor="let c of filteredData; let cIndex = index">
 						<!-- Col 1, Citation -->
-						<td class="width-60">
+						<td class="width-60" (click)= "setSelectedCitation(c); d('closed')">
 							<div class="half-space">
 								<span *ngIf="!c.isEditing">{{ c.author }}</span>
 								<span *ngIf="c.isEditing"><textarea class='break' name="author" [(ngModel)]="c.author" required></textarea></span>
@@ -96,11 +96,11 @@
 							</div>
 						</td>
 						<!-- Col 3, Associated Regression Region and Buttons -->
-						<td class="width-35">
+						<td class="width-35" *ngIf="!inGagePage">
 							<div *ngFor="let e of c.regressionRegions; let lst = last;">{{e.name}}<span *ngIf = "!lst"><hr></span></div>
 						</td>
 						<!-- Col 4 - Regression Buttons -->
-						<td class="width-5">
+						<td class="width-5" *ngIf="!inGagePage">
 							<div *ngFor="let e of c.regressionRegions; let lst = last;" [class.hidden]="!showAddCitations" class="cell-buttons">
 								<button title="Edit Regression Region" (click)="editRegRegion(e.id)" class="icon-button" [disabled]="!checkManagerRegressionRegions(e)" aria-hidden="true">
 									<i class="far fa-pencil"></i>
--- a/src/app/shared/components/managecitations/managecitations.component.html
+++ b/src/app/shared/components/managecitations/managecitations.component.html
@@ -47,7 +47,7 @@
 			</form>
 
 			
-			<table class="custom-table table-striped table-scroll mtop-sm no-scrollbars" [class.table-hover-rows]="inGagePage">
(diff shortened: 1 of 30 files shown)


kjacobsen16:
@Anders-Hopkins I tried to pull this down and test it with one of the test stations, but I'm getting errors trying to edit or add characteristics or stats

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -50,87 +50,148 @@
...
 					<tr>
-						<th class=citationHeader>Citation</th>
-						<th class=citationEditHeader></th>
-						<th class=regRegionHeader>Associated Regression Regions</th>
-						<th class="empty-cell regRegionEditHeader">
+						<th *ngIf="!inGagePage" class=citationHeader>Citation</th>
+						<th *ngIf="!inGagePage" class=citationEditHeader></th>
+						<th *ngIf="inGagePage" class="width-5">ID</th>
+						<th *ngIf="inGagePage" class="width-85">Citation</th>
+						<th *ngIf="inGagePage" class="width-10"></th>
```
kjacobsen16: It looks like some of these are repeats, i.e. the citationHeader and citationEditHeader. Those two should be in there regardless. You can conditionally set the class if you need to, but otherwise just adding the ID here and keeping the others should be fine.

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -50,87 +50,148 @@
...
+										<i class="far fa-times"></i>
+									</button>
+								</div>
+								<!-- Not editing -->
+								<div [class.hidden]="showAddCitations" class="cell-buttons">
+									<button title="Add Citation" (click)="addExistingCitation(c); d('closed')"
+										class="icon-button" aria-hidden="true">
+										<i class="far fa-plus"></i>
+									</button>
+								</div>
```
kjacobsen16: This is the section that allows a user to select a citation. This happens when a user in NSS is in the add regression region modal and hits "Find Existing Citation". You should be able to use this here and then just handle the click slightly differently.

src/app/shared/components/managecitations/managecitations.component.html
```diff
@@ -50,87 +50,148 @@
...
+									<span *ngIf="c.isEditing"><textarea class='break' name="title" [(ngModel)]="c.title"
+											required></textarea></span>
+								</div>
+								<div>
+									<a *ngIf="!c.isEditing" [href]="c.citationURL" [target]="_blank">
+										<span *ngIf="!c.isEditing">{{ c.citationURL }}</span>
+									</a>
+									<span *ngIf="c.isEditing"><textarea class='break' name="citationURL"
+											[(ngModel)]="c.citationURL" required></textarea></span>
+								</div>
```
kjacobsen16: Again, this section and on has a lot of repeats that can be treated more or less the same if it's opened from gage page or not.

kjacobsen16 (changes requested):
Overall, a big improvement! A couple things I think would be helpful:
1. Putting a button in the main GageStats page to open the manage citation modal, should be pretty easy since it's already set up (can be a separate issue if you want)
2. This is a little confusing, but gagestats and nss have different citation tables, so the citations calls in the gagestats portion should actually use the `gagestatsservices/citations` endpoint
3. Can you make it more clear that clicking on a citation will select it? It took me a while to try just clicking on it since the cursor didn't change or anything. NSS has the "+" icon that you click to select it.
4. I can't see the full text for the "Add Citation" button. You could use an icon instead (like a magnifying glass or a list icon), or we can try to make the modal a bit bigger again.
image

src/app/shared/components/managecitations/managecitations.component.html
Anders-Hopkins: Yea, so the reason why there are so many repeats here and especially the repeated columns below is because there needs to be different styling depending on whether the Citation Modal is being opened in a GagePage or from the MainView. Is there a way to make the CSS class dependent upon a Boolean variable? For instance; ngIf="inGagePage=true" class = A, else class = B.

kjacobsen16: Yes, you can conditionally set the class by doing something like `[class.width-5]="inGagePage"`

kjacobsen16: Another thought is you could set a conditional class higher up, something like `[class.gageStatsTable]="inGagePage"`, then use the existing classes here (citationHeader, etc.) combined with that class in the scss file, e.g. `.gageStatsTable.citationHeader` (it might be different for scss v. css)

kjacobsen16 (changes requested):
Sorry to keep coming back with edits! There are more requests that need to go to the gagestats url instead. Namely, adding, editing and deleting citations in the manage citations modal hits nss instead of gagestats in the gagestats page (honestly at some point we should just move it to a shared table but that's far off).

We might want to consider changing how our settings service is doing the post/gagestatspost, etc. stuff, and instead of using the base url in the settings service, send it through in the component that's calling the request. That would require changing every instance that's calling them, but should be pretty simple. Let me know if that doesn't make sense!

kjacobsen16 (approved):
Looks good - I'm adding a new issue to update the list of citations in the gagestats manage citations modal when there's an update to a citation (post, put, or delete)

PR #363, opened 2020-11-03
Fixes #362

### LINKED ISSUE #362: Unit Types Settings Page - Show Unit System Name instead of ID
image
Instead of showing the unit system id, we should display the unit system name

harper-wavra:
Displays unit system name instead of ID.


kjacobsen16 (approved):
👍

PR #365, opened 2020-11-09
Fixes #325

### LINKED ISSUE #325: Edit/Add/Delete Methods in Settings
We should also be able to edit/add/delete methods in the settings page waiting for service issue [#175](https://github.com/USGS-WiM/NSSServices/issues/175).

harper-wavra:
Able to add/delete/edit methods, as well as "refresh" methods when adding new regression region.


kjacobsen16 (approved):
Works great!

PR #366, opened 2020-11-12
Fixes #334

### LINKED ISSUE #334: Add scenario modal changes - Expected values
Edit: see my comment below (we still might want to consider what I wrote out here but I think the skipCheck option will be the easiest to enact).

Writing out my thoughts here, but might need more thought.

Pete and Ryan have requested to remove the "Expected" portion of the equations and instead have some sort of equation test because users won't always have example values. 

We could change the services, but I had a thought that might work for just the client: we could maybe just use the max/min of each parameter and plug it into the equation client side to send as the expected value. In this case, we could just auto-fill in the example values based on that, then it gives the user flexibility if they _do_ have example values, and if for some reason the services calculated differently than the client, the user can then adjust the expected value themselves.

harper-wavra:
Added "skipCheck=true" to the url to skip the equation values comparison

### DIFF
--- a/src/app/shared/components/addscenario/addscenario.component.ts
+++ b/src/app/shared/components/addscenario/addscenario.component.ts
@@ -338,23 +338,25 @@ export class AddScenarioModal implements OnInit, OnDestroy {
         this.fillModal();
     }
 
-    public cloneOrEdit(){
+    public cloneOrEdit(e){
         //make sure in edit mode
-        if (this.editMode == true){
-            this.scen = JSON.parse(JSON.stringify(this.newScenForm.value));
-            //change back to edit if user reselects original core dropdowns
-            if ((this.originalScenario[0] == this.scen.statisticGroupID) &&
-            (this.originalScenario[1] == this.scen.regressionRegions.ID) && 
-            (this.originalScenario[2] == this.scen.regressionRegions.regressions.ID)){
-                this.clone = false;
-                this.edit = true;
-                this._toasterService.clear();
-                this._toasterService.pop('info', 'Info', 'Scenario Will Be Edited Instead Of Cloned');
-            }else{ //change to clone
-                this.clone = true;
-                this.edit = false;
-                this._toasterService.clear();
-                this._toasterService.pop('info', 'Info', 'Scenario Will Be Cloned Instead Of Edited');
+        if(e){
+            if (this.editMode == true){
+                this.scen = JSON.parse(JSON.stringify(this.newScenForm.value));
+                //change back to edit if user reselects original core dropdowns
+                if ((this.originalScenario[0] == this.scen.statisticGroupID) &&
+                (this.originalScenario[1] == this.scen.regressionRegions.ID) && 
+                (this.originalScenario[2] == this.scen.regressionRegions.regressions.ID)) {
+                    this.clone = false;
+                    this.edit = true;
+                    this._toasterService.clear();
+                    this._toasterService.pop('info', 'Info', 'Scenario Will Be Edited Instead Of Cloned');
+                } else { //change to clone
+                    this.clone = true;
+                    this.edit = false;
+                    this._toasterService.clear();
+                    this._toasterService.pop('info', 'Info', 'Scenario Will Be Cloned Instead Of Edited');
+                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -168,9 +170,30 @@ export class AddScenarioModal implements OnInit, OnDestroy {
         });
     }
 
+    public equationCheck(check){
+        if (check == true) {
```
kjacobsen16: Not a mandatory change, but since `check` is a boolean, you can just say `if (check) {` here

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -168,9 +170,30 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
+            this.skipCheck = true;
```
kjacobsen16: Same thing as above, because `check` is a boolean, you can actually set `this.skipCheck = check;` at the beginning of the function.

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -168,9 +170,30 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
+            this.newScenForm.get('regressionRegions.regressions.expected.value').disable();            
+            this.newScenForm.patchValue({ regressionRegions: { regressions: { expected: { value: null}}}});
+            const parmControl = <FormArray>this.newScenForm.get('regressionRegions.parameters');
+            for (let i = parmControl.length-1; i >= 0; i--) {
+                this.newScenForm.get('regressionRegions.parameters.'+i+'.value').disable();
+                parmControl.controls[i].get('value').setValue(null);
+            }
```
kjacobsen16: So you can actually set this in the html, I think it's like `[disabled]="skipCheck"`, then this whole function can be removed and you can just change the skipcheck in the html instead.

src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -445,7 +472,11 @@ export class AddScenarioModal implements OnInit, OnDestroy {
         // add parameter name, description, add values/check if between limits
         regs.expected.parameters = {};
         for (const parameter of regRegs.parameters) {
+            if (this.skipCheck == true) {
+                regs.expected.parameters[parameter.code] = 0;
+            } else {
```
kjacobsen16, line 480: This is fine, but if you're skipping the equation check it doesn't really matter if you have values or not, so you don't actually have to go through and remove these :)

kjacobsen16 (changes requested):
Looks good! Only real issue is that we should also disable the expected prediction interval upper and lower bounds when the user is skipping the equation check.

The rest are all just cleanup suggestions.

src/app/shared/components/addscenario/addscenario.component.ts
harper-wavra: If I didn't set a value for the parameters it would tell me the equation was invalid. I tried in Insomnia to not have a value for the parameters and it wouldn't go through. I included the ways I tried and if they worked or not.
[works](https://user-images.githubusercontent.com/59063181/99092610-fb843600-2596-11eb-8025-dc2689ba2224.png)
[doesn't work](https://user-images.githubusercontent.com/59063181/99092751-2a9aa780-2597-11eb-82c3-3fbe4d4436dd.png)
[doesn't work](https://user-images.githubusercontent.com/59063181/99093004-8c5b1180-2597-11eb-9fff-ff6a8398c4b8.png)

kjacobsen16: OK, gotcha!

src/app/shared/components/addscenario/addscenario.component.ts
harper-wavra: I tried to get this all to work in the html but I wasn't able too. I researched it a little and it seems you are unable use [disabled] within Reactive Forms, and instead are suppose to use the disable() method from the FormControl, but maybe I am misunderstanding?

kjacobsen16: Oh, interesting. I thought we were using it somewhere in the app. Maybe we were using a disabled class instead? `[class.disabled]="skipCheck"`

kjacobsen16: Or it could be that those aren't working/aren't in forms?

harper-wavra: We use [disabled] for buttons so I think those works because they are not in the forms, maybe.

kjacobsen16 (approved):

Issue #367, opened 2020-11-16
Update GageStats Citations List on Post/Put/Delete


kjacobsen16:
The gagestats manage citations modal isn't showing updates to the citations until you close and reopen the modal.

This is because the new citations are triggered by resetting the selecting region, because that would reset it as far as the NSS side goes. That shouldn't update the gagestats citations list, and I also noticed one time that it actually set the selected region, so when I went back to NSS it looked like one was selected.

Issue #360, opened 2020-10-29
Set timer on GagePage Toastr pop-ups


Anders-Hopkins:
Toastr pop-ups are produced when new characteristics and statistics are added to a station gage. These pop-ups only disappear when the user clicks on them. This would be better if they disappear on a users' click or after 3-5 seconds.

Anders-Hopkins:
Ok, this is waaaaaaay simpler than I thought it would be. Probably didn't need its own issue, but then I thought there would be more changes required to do this.

kjacobsen16:
@Anders-Hopkins did this get pushed through with that last pull request?

Anders-Hopkins:
@kjacobsen16  Yes!

kjacobsen16:
Thanks! I just tested it and noticed that there are a few more places where a `ToasterConfig` is made that need it, should be an easy fix if someone wants to throw it in with another pull request

PR #373, opened 2020-11-17
Solves #367 & #360

Anders-Hopkins:
Set 5 sec timers on all Toast pop-ups. Also set the citations to reload after a http request. I hah difficultly replicating the problems discussed in #367 so I hope this solves the issue as needed.


kjacobsen16 (approved):
Looks good!

PR #378, opened 2020-11-30
Fixes #372

### LINKED ISSUE #372: Edit Scenario Headers
When you click 'Edit Scenario' the headers can be a little confusing. The idea for the headers was that the statistic group is a header, then each regression region is a "sub" header within that statistic group So groupings are like the scenarios where stat group > regression regions > citation, parameters, equations.

- [ ] Status and Method should be pulled out so the region header doesn't seem so long, maybe put it below it with different formatting (smaller text, non-bold, etc.)?

- [ ] Play with styling (maybe ask @mitchas ), so that it's more obvious that both the Statistic group and regression region are their own headers, but that they are different. I think we used to have some borders on the bottom with difference thicknesses

[image](https://user-images.githubusercontent.com/59063181/99409536-1d432d00-28b7-11eb-9af5-074d0b80886a.png)

harper-wavra:
I think this looks better, let me know what you think.

### DIFF
--- a/src/app/mainview/mainview.component.html
+++ b/src/app/mainview/mainview.component.html
@@ -928,7 +928,7 @@
 
 	<!-- editRegionScenario -->
 	<div class="hidden-print" *ngIf="editRegionScenario">
-		<form #editScenarioForm="ngForm">
+		<form #editScenarioForm="ngForm" class="editScen">
 			<div>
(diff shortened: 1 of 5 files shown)


kjacobsen16 (approved):
This looks great, thanks @harper-wavra and @mitchas!

Issue #206, opened 2020-05-27
General cleanup checklist


kjacobsen16:
As we go, I notice general cleanup that can be done. Putting a spot here for areas that I notice should be addressed but we don't have time for now

- [ ] add regression region modal `showNewRegressionRegionForm()` function - see if we can find a better way to set each part of the form based on the object.

- [ ] Imports, formatting, comments?

- [ ] Metadata at top of some files (e.g. addscenario component ts file) - most of it was copied when I started the project, so the person at the top doesn't actually mean they created it

- [ ] Add regression region modal - edit citation functionality

PR #381, opened 2020-12-04
Fixes #263

### LINKED ISSUE #263: Add Limitations functionality to NSS
Not an enhancement, but less important than other issues.

- [x] Display limitations for regression regions - may be fine to just display in edit regression region modal?
    - comes back when you get regression region by ID, but can also get [here](https://test.streamstats.usgs.gov/docs/nssservices/#/Limitations/GET/RegressionRegionLimitations).

- [x] Allow user to add limitations to a regression region - you can just do this by attaching a "Limitations" object to the regression region. More info [here](https://test.streamstats.usgs.gov/docs/nssservices/#/Limitations/POST/AddRegressionRegionLimitations).

- [x] Allow user to edit limitations - this can also be done by modifying the limitations from the regression region by id endpoint (currently used by the edit reg region modal) and sending it through with the edited regression region.

- [x] Allow user to delete limitations from regression regions. This has to be done by hitting the limitations endpoint directly with the ID of the limitation to be deleted. More info [here](https://test.streamstats.usgs.gov/docs/nssservices/#/Limitations/DELETE/DeleteLimitation).

Waiting on issue - https://github.com/USGS-WiM/NSSServices/issues/186

harper-wavra:
I made it so it works sort of like the editing of gages. The edit limitations section is sort of its own section and the changes made to the limitations get saved right away. 

Files editied:
addregressionregion.component.html - added limitations form and spacing
addregressionregion.component.scss - added table styling
addregressionregion.component.ts - limitation code and organized into sections
managecitations.component.html - spacing
managecitations.component.ts - remove console.log
app.service.ts - spacing
config - added limitations URL


src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -493,5 +548,152 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
+        if (unit.id.toString() == v.unitTypeID.toString()){
+          control.controls[varIndex].get('unitTypeID').setValue(this.unitTypes[x].id);
+        }
+      });
+      this.variables.forEach((variable,x) => {  
+        if (variable.id.toString() == v.variableTypeID.toString()){
+          control.controls[varIndex].get('variableTypeID').setValue(this.variables[x].id);
+        }
+      });
+    });
```
kjacobsen16, line 637: Is this how we update items in the add scenario modal? I wonder if there's a better way to do this by linking the html directly with it? If it looks like a long task, we could maybe add it to #206 (which I haven't looked at in a while)

src/app/shared/components/addregressionregion/addregressionregion.component.ts
```diff
@@ -493,5 +548,152 @@ export class AddRegressionRegionModal implements OnInit, OnDestroy {
...
+        this._toasterService.pop('error', 'Error creating Limitations', error.message || error._body.message || error.statusText);
+      }
+    );  
+  }
+
+  public formatLimitation() {
+    if (this.newLimForm.value.regressionRegionID) {
+      this.newLimitation = []; 
+      this.newLimitation.push(this.newLimForm.value);
+      this.createNewLimitation(this.newLimitation);
```
kjacobsen16, line 688: We might be able to tie the limitations in with the reg region form instead, then we wouldn't need to worry about storing these and adding them separately. I think the add scenario component does that with the parameters and errors?

kjacobsen16 (changes requested):
Overall, looking good! Thanks for the breakdown of your changes, that saved me some time. A couple notes that can wait until the rest of the non-management side issues are fixed:
- [x]  Can we limit the variable units for the limitations to the units assigned to the variable type, similarly to how we're limiting them now in the add scenario modal?
- [ ] I think I deleted a limitation and it deleted before I hit save.. Can you double check that and make sure all changes are going through after the user saves the regression region?
- [x] While we're in the add regression region modal, can you make the "Method" field optional for now (if possible)? and also move the "Code" input up to the top, before Status?

harper-wavra:
This is in reply to your second comment. So I wasn't able to make all changes  go through after the user saves the regression region because there were too many things the user could add or edit or delete before saving. We have the citation and location saving after the user saves the regression region because they user can only create or edit or delete that one item. For the limitations however, the user can add or edit or delete each limitation and also add or edit or delete each variable within each limitation (maybe I am making it too complicated in my head though.) So I made it work like it works when you are editing a gages characteristics or statistics. The limitations section is sort of its own section and the changes made to the limitations get saved right away. That is why I created that "Edit Limitations" button inside the regression modal to try and make it clear to the user your changes to limitations are getting saved right away.

harper-wavra:
Waiting on service issue https://github.com/USGS-WiM/NSSServices/issues/186(https://github.com/USGS-WiM/NSSServices/issues/186)

PR #383, opened 2020-12-10
Fixes #364

### LINKED ISSUE #364: Updates from NSS Service Changes
I'm including all of these here because there are some commonalities, we may want to break them out a bit further.

The variable type's defaultUnitTypeID changed to two fields: metricUnitTypeID and englishUnitTypeID

- [x] display this change in the settings page table

- [x] allow user to select both when adding/editing a variable type

- [x] In the add scenario modal, Ryan wants the users limited to selecting either the metric or english unit type when selecting the unit type for each scenario parameter - note that sometimes the english and metric will be the same (i.e. dimensionless)

Also, now the variable types have statistic groups associated with them.

- [x] Display the statistic group for each variable type in the settings page

- [x] allow the selection of a statistic group when adding/editing a variable type - note that statistic groups now have a "def type" (this was a term from SSDB to define if it's used for statistics or basin characteristics), they should only be allowed to choose stat groups with the "BC" def type

Related changes:

- [x] We should limit the statistic group selection in the add scenario modal to statistic groups with the "FS" def type

- [x] Gage Page - display the statistic groups for the basin characteristics, like we do with the statistics. Note that in this case, the statistic group is assigned to the variable type, not the characteristic like how the statistic is done.

harper-wavra:
- Display variables metricUnitTypeID and englishUnitTypeID in the settings page table and allow user to select both when adding/editing a variable type.
- Users are limited to selecting either the metric or english unit type when selecting the unit type for each scenario parameter in add scenario modal.
- Display the statistic group for each variable type in the settings page and allow the selection of a statistic group when adding/editing a variable type (only "BC" def type)
-  limit the statistic group selection in the add scenario modal to statistic groups with the "FS" def type
- displays and filters the statistic groups for the basin characteristics

Files Changes
- gagepage.component.html - added html to filter and show the characteristics statistic groups
- gagepage.component.ts - added code to filter and sort the characteristics into statistic groups
- variabletypes.component.html - displays english and metric unit type, allows user to select both when adding/editing a variable type
- variabletypes.component.ts - code to allow user to select english and metric unit type when adding/editing a variable type
- addscenario.component.html - display either the metric or english unit type when selecting the unit type for each scenario parameter
- addscenario.component.ts - code to limit users to selecting either the metric or english unit type when selecting the unit type for each scenario parameter
- not sure why the dist files changed


src/app/shared/components/addscenario/addscenario.component.ts
```diff
@@ -239,16 +240,32 @@ export class AddScenarioModal implements OnInit, OnDestroy {
...
-            controlArray.controls[varIndex].get('unitType').setValue(null);
+        }
+        if (this.variables.find(r => r.code === varCode).metricUnitTypeID) {
+            this.unitTypes.forEach((element,index) => {  
+                if (element.id.toString() == this.variables.find(r => r.code === varCode).metricUnitTypeID) {
+                    if (tempUnit!= this.unitTypes[index]) { // Checking for duplicates
+                        this.defaultUnitTypes.push(this.unitTypes[index]);
+                    }
+                    findUnits = true;
+                }
```
kjacobsen16: The whole MetricUnitType and EnglishUnitType are both included in the variables objects, so you don't actually need to look through the `this.unitTypes` object here at all, just add the two default units to the list (and check that the default unit IDs are different before adding the metric unit type)

kjacobsen16 (changes requested):
Great job, lots of important changes here. A few things:
- [x]  When a new variable type for a parameter is changed, we should clear the unit associated with it in case the default units change.
- [x] The dropdown gets cut off when filtering the physical characteristics in the gage page - Mitch might be able to help here.
image
- [x] This is my bad, but we should also limit the unit selection when adding characteristics in the gage page to the default variable types. Sorry about that!
- [x] Again, something I missed - can you display the DefType in the Statistic Groups Table in the Settings page and allow assigning the deftype when creating a new statistic group? Options should be "FS" (meaning Flow Statistics) or "BC" (meaning Basic Characteristics).

Feel free to make a separate issue for some of these.

harper-wavra:
I created new issues for the last two changes requested. 

@mitch-wim could you take a look at the physical characteristic "Filter by StatisticGroup" dropdown in the Gagepage (the second comment)? 50276000 is one gage that has this table.

kjacobsen16 (approved):
👍

PR #384, opened 2020-12-11
Fixes #374

### LINKED ISSUE #374: Regions do not load in GageStats table
The regions which are used to populate the 'Study Area' column in the GageStats table do not load when the GageStats page is opened from another page in NSS. If the GageStats page is refreshed, then the 'Study Area' column populates. This must have something to do with the asynchronous nature of the get request for regions.

harper-wavra:
Loads regions when coming to Gagestats from NSS page.


kjacobsen16 (approved):
👍

PR #391, opened 2020-12-16
Fixes #241

### LINKED ISSUE #241: Weird Issue displaying errors
I think because the first regression in the second section doesn't have errors, it's not creating the columns correctly
image

To replicate, compute scenarios for the test region

harper-wavra:
This is a fix for the errors displaying weird (I think this was only issue in the test region.) I looked at a couple other regions and the errors still seem to be displaying correctly.


kjacobsen16 (changes requested):
Looking good! It just looks a little weird when there are no errors for a regression (see the KJTest region)
image

kjacobsen16 (approved):
👍

PR #396, opened 2020-12-29
Fixes #394

### LINKED ISSUE #394: Make /gagestats route open to the public
There must be a routing constraint on the /gagestats route. I get rerouted from https://test.streamstats.usgs.gov/nss-dev/gagestats to https://test.streamstats.usgs.gov/nss-dev/ if I'm not logged in. Users should be able to use the /gagestats url regardless of being logged in.

harper-wavra:
Removed unused `returnUrl ` variable. 
If path is gageStats doesn't reroute to NSS.


kjacobsen16 (approved):
Works great!

Issue #379, opened 2020-12-01
Edit gagepage console errors


harper-wavra:
When you are finishing adding a statistic or characteristic to gage and click the save button I get the following console errors.
![image](https://user-images.githubusercontent.com/59063181/100766439-41763200-33be-11eb-858c-d54500080277.png)

harper-wavra:
Fixed with https://github.com/USGS-WiM/NSS/pull/400

PR #400, opened 2021-01-05
Fixes #387 and #379

### LINKED ISSUE #387: GagePage - Limit Unit Selection
We should limit the unit selection when adding characteristics in the gage page to the default variable types.

harper-wavra:
- Limited the unit selection when adding a characteristic
- Removed printing values of `newChar` and `newStat` in table while waiting for the services and instead added a loader (which also got rid of the console errors when adding a new characteristic.)
- Was able to remove console errors when adding a new statistic by changing `ngIf` slightly.


kjacobsen16 (approved):
Looks good!

PR #401, opened 2021-01-08
Fixes #319

### LINKED ISSUE #319: Add Station Updates
- [x] Specify that lat/long should be location _in stream_

- [x] Add field for lat/long sources - this will have to wait on https://github.com/USGS-WiM/GageStatsServices/issues/75

harper-wavra:
Added 'in stream' text after 'Latitude' and 'Longitude' in both the gagepage and add station modal to specify that the location should be in stream.

Added required field in add station modal for location source.

Displays location source in gagepage and allows users to edit it.

### DIFF
--- a/src/app/gagestats/add-station/add-station.component.html
+++ b/src/app/gagestats/add-station/add-station.component.html
@@ -55,10 +55,9 @@
                         is required</div>
                 </div>
 
-                
                 <!-- latitude -->
                 <div class="form-group required width-40">
-                    <label class="req" for="latitude">Latitude:</label>
+                    <label class="req" for="latitude">Latitude (in stream):</label>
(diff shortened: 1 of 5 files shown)


kjacobsen16 (approved):
Looks good!

I'm noticing some strange issues when saving the top section of a gage when it also has statistics and characteristics (e.g. station 01239999). It's a weird service issue I'll look into, but I think if we remove the statistics and characteristics objects from the station when sending through the edits to the top section, that should fix it on the client side. We really only need the following basic info we get from the /stations/ call when editing the non-statistic/characteristic parts of the gage:
```
{
  "id": 36541,
  "code": "01239999",
  "agencyID": 3,
  "name": "test station",
  "isRegulated": false,
  "stationTypeID": 8,
  "location": {
    "type": "Point",
    "coordinates": [
      44.444,
      -99.999
    ]
  },
  "locationSource": "test source",
  "regionID": 76
}
```

harper-wavra:
I think that last commit should have fixed it.

PR #406, opened 2021-01-11
Fixes #138

### LINKED ISSUE #138: Bulk Upload
Add button in order to do Bulk Upload of new stations. 

Might need to do research on deciding the best way to bulk upload.
- upload cvs?
- copy in table?

thumbnail

Anders-Hopkins:
Added a modal for the bulk upload of stations, stats, and chars. The modal is fully functional, but perhaps could benefit from some aesthetic changes.


mitchas:
I adjusted the styles and layout a bit. Is there an example file I could upload to populate the table correctly to make sure it looks fine?

Anders-Hopkins:
Here is the excel file I have been using to develop this modal. Bear in mind that some of the values in the spreadsheet are erroneous and will be flagged were you verify the data before submission.
[example_gagestats_bulk_upload.xlsx](https://github.com/USGS-WiM/NSS/files/5814823/example_gagestats_bulk_upload.xlsx)

src/app/gagestats/batch-upload/batch-upload.component.html
```diff
@@ -0,0 +1,110 @@
...
+                <ng-container *ngFor="let sheetName of wb.SheetNames">
+                        <button class="button grey mright-sm mtop-xs" (click)='selectSheet(sheetName)'>{{sheetName}}</button>
+                </ng-container>
+			</div>
+
+			<!-- Edit buttons -->
+			<div class="edit-buttons">
+				<button *ngIf='tableDisplay && !tableEdit' type='button' class="button small grey mright-xs" (click)='tableEdit = true && disableSumbit = true'>Edit Table</button>
+                <button *ngIf='tableDisplay && tableEdit' type='button' class="button small grey mright-xs" (click)='tableEdit = false'>Finish Edits</button>
+                <button *ngIf='tableDisplay && !tableEdit' type='button' class="button small grey mright-xs" (click)='showManageCitationsModal()'>Add Citation</button>
```
kjacobsen16: This is needed for statistics and characteristics, but not stations (the add citation button)

src/app/gagestats/gagestats.component.ts
```diff
@@ -147,7 +147,8 @@ export class GagestatsComponent implements OnInit {
   }
 
   //TODO: Bulk Upload Button
```
kjacobsen16: remove TODO here

src/app/gagestats/batch-upload/batch-upload.component.ts
```diff
@@ -0,0 +1,438 @@
...
+            }
+            if(recordObj.isPreferred) {
+              var x = this.getTrueFalse(recordObj.isPreferred);
+              recordObj.isPreferred = x;
+            }
+            if(recordObj.code) {
+              var cellIndex = Object.keys(recordObj).indexOf('code');
+              this.getStationID(recordObj, rowID, cellIndex);
+            }
+            recordObj.comments = 'Statistic Date Range: ' + recordObj.startDate + ' - ' + recordObj.endDate + '.';
```
kjacobsen16: (line 295) Might have issues if startdate or enddate aren't given

kjacobsen16 (changes requested):
Great job! This is working really well over all. It's a big change, so I have a handful of comments:

**Verifying data**:
1. I forgot that it didn't automatically pick the corresponding property name for each column, so I didn't add those at first. When I hit "Verify data" it still thought everything was right. We should fix that and think about ways to make it more obvious to the user and notify them when they're missing.
2. Can we check if any of the column names from the spreadsheet match the required properties and assign them on load? Users would still be able to change them and assign those that don't match anything, but it would be more convenient if those were already selected where possible.
3. It may be helpful to consider using [angular forms](https://angular.io/api/forms). We use those for the Settings page to validate the data entered into the Add modals. We do that by creating a formgroup that has all of the properties we'd expect, default values (if desired) and whether they're required. There's also validation built into forms that we can use in the HTML (for an example, try typing in a field in one of the settings add modals and deleting it, it should notify you it's required). It may be a little difficult with how the excel import works, but may be worth looking into to keep consistent with the rest of the app. In those cases, we usually hard code each property in the HTML, but I think we could automate it by looping through the form group properties...? Just a thought.
4. When editing data that's incorrect, it's hard to know what _is_ correct. Some thoughts on this are: (1) add dropdowns of things where we can, e.g. allow a user to choose a variable type from a dropdown, (2) think about adding tooltips for additional information. Number 2 may be something we add as a separate issue as it would require extra planning and brainstorming of the terminology
5.  When adding stations, we should check to make sure they don't already exist.

**General notes**: 
1. When we get a response back after submitting changes, we should requery the main gagestats filters. I added some stations and it didn't update the number of pages, although it went from 730 pages to 731. Same when gages are deleted, but we can add a separate issue for that unless you find it easy to add in.
2. I had a strange issue where I selected a property name for a column, then accidentally hit the "x" next to it, then I couldn't select it again until I selected something else first.
3. Can we enable adding a line in the table? This could be another enhancement issue to add later.
4.  We've changed the variable types a bit in that they all have metric and english units (and will be doing the same for regression types at some point). For variable types for now, we should only allow them to assign a unit type that is assigned as either the metric or english unit type for that variable type.
5. If "Remarks" are null, we should not append them to the Comments property. It's showing up as "Statistic Date Range: 07/28/1960 - 07/28/2020. null" Also, in other Statistics it goes "Statistic Date Range: ... Other Remarks ..."

That may seem like a lot, but most of them are just enhancing what you're already doing. This really is working and looking great! A lot of these could probably be added as separate issues or enhancements, so feel free to ask me if you're unsure about prioritization.

src/app/gagestats/gagestats.component.ts
Anders-Hopkins: Removed

Anders-Hopkins:
Response to suggested edits:

**Verifying Data:**
1. This got fixed.
2. I set it so that if the header of the Excel column matches that of the 'id' of 'dropdownOptions' array. The columns that do not get matched will be highlighted in the verify data function.
3. Using Angular Forms is probably a good idea. I didn't get to this since I don't fully understand them.
4. I tried to get dropdown menus to appear in the 'edit table' for the values which are erroneous. This is pretty tricky and I didn't get it to work  and probably spent too much time on this.
5. Didn't get to this yet.

**General Notes:**
1.  I didn't get to this.
2.  I know of this issue. If you hit the 'delete' key instead of the 'x' button, this clears the selected contents of the dropdown menu and adds this selected value back to the menu. This is a bug with the dropdown menu tool that I have spent a good deal of time trying to fix but haven't yet.
3.  I have laid the groundwork for this but it still needs work. An example is when adding stations, clicking 'Add new row' adds a new row but the cell for 'IsRegulated' adds a regular cell not a checkbox.
4.  I didn't understand this part. Are there separate metric units and english units lists?
5.  Working on this too. Having an issue getting it to work. `if(recordObj.remarks ===  null)` never returns `true`. Not sure what is going on here.

kjacobsen16:
@Anders-Hopkins to answer your question `Are there separate metric units and english units lists?`: all of the units are in a combined "units" table and endpoint, but we assigned a "metric unit type" and an "english unit type" from the units table to each variable type. It's also now a step when adding a new variable type, the user must select a metric unit type and english unit type. Because they are all combined in one table, it's up to the user to make sure they select the right unit for each of those.

I'm adding a comment for that item in #425

kjacobsen16 (approved):
I believe we now have all outstanding fixes in issues and the general functionality is working well. 👍  thanks @Anders-Hopkins!

PR #407, opened 2021-01-12
Fixes #405

### LINKED ISSUE #405: Add study area to gage page
Display study area (region) in gage page and allow editing like the rest of the "Gage Information".

harper-wavra:
Able to view and edit region in gage page.

I noticed when editing the region on gages that sometimes you are changing the region for only gage with that certain region (or station type, agency, ect) which in turn makes the sidebar dropdowns incorrect. Therefor, I added code to requery the GageStats sidebar when a gage is edited, deleted, or added.

### DIFF
--- a/src/app/gagestats/add-station/add-station.component.ts
+++ b/src/app/gagestats/add-station/add-station.component.ts
@@ -105,6 +105,7 @@ export class AddStationModal implements OnInit {
         }
         this.cancelSubmitStation();
         this._nssService.searchStations(this.selectedParams);
+        this._nssService.setRequryGSFilter(true);
       }, error => {
         if (!this._settingsService.outputWimMessages(error)) {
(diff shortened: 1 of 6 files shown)


kjacobsen16 (commented):
A couple changes:
- [x] Some typos - some of the `requery` variables are spelt `requry`
- [x] In the gage page, "Region" should be "Study Area" to keep consistent with other places
- [x] If you requery the sidebar and the region is no longer part of the returns (e.g. if I change the last station for my current filters from Test to Undefined), it shows up like the following image. This is an edge case, so we can make this an enhancement issue if you don't see an immediate fix.
image

harper-wavra:
@kjacobsen16  The first two comments should now be fixed and I created a new issue for the third.

kjacobsen16 (approved):
👍

PR #411, opened 2021-01-26
Fixes #370

### LINKED ISSUE #370: Gagestats Lat Long
We need to have NWIS lat/long and StreamStats lat/long.
NWIS lat/long would come from NWIS and StreamStats lat/long would be defined as gage locations snapped to the flow accumulation grid that is submitted by WSCs. This recently came up in a discussion with a WSC and they are not able to change the NWIS location, but want the gage plotted on the flow accumulation grid. Thus we need a lat/long for both. 

From Pete's Feedback 10/29/20

harper-wavra:
Added NWIS lat/long to gagepage.
Displays N/A for gages without a waterservice page (ie those in Ontario)


kjacobsen16 (approved):
Looks good, I made some tiny changes to the display text

PR #412, opened 2021-02-02
Fixes #399

### LINKED ISSUE #399: Rethink Settings Filters
Filters like the one shown below in the Statistic Groups settings page might not work accurately due to the statistic groups being used in both GageStats and NSS. I don't think either services are set up to check both databases. We would need to either (a) remove these filters or (b) query both services (NSS and GageStats) with the associated filters and combine the results.
image

harper-wavra:
In the settings for regression types and statistic groups (since they are the only lookups with filters) I queried both services (NSS and GageStats) when the filters were selected and combined the results. 

- regressiontypes.component.ts - queried both services and combined results for regressionTypes, added loader, removed spacing, renamed variable, removed unused variables
- regressiontypes.component.html - renamed variable
- statisticgroups.component.ts - queried both services and combined results for statistic groups, removed spacing, removed unused variables
- agencies, errors, managers, regions, stationtypes, unitsystems, unittypes, variabletypes component.ts - removed spacing between imports and removed 3 unused variables that must have been copied from a different component when these were created. 
- dist - changed because I updated dev site


src/app/settings/categories/regressiontypes/regressiontypes.component.ts
```diff
@@ -99,13 +98,19 @@ export class RegressionTypesComponent implements OnInit, OnDestroy {
...
-            .subscribe(regs => {
-                this.regressionTypes = regs;
+            .subscribe(res => {
+                this.nssRegressionTypes = res;
+            });
+        this._settingsservice
+            .getEntities(this.configSettings.gageStatsBaseURL + this.configSettings.regTypeURL+"?regions="+ this.selectedRegionID +"&statisticgroups="+ this.selectedStatisticID)
+            .subscribe(res => {
+                this.gsRegressionTypes = res;
+                this.combineRegressionTypes();
```
kjacobsen16: So, my only thought here is that this might cause issues if the gagestats service calls come back before the nss calls. In that case, we might not be combining things correctly. What we should do here is have a check for whether the other response has come back and if it has then call `combineRegressionTypes`. We'll want to do that when we call the nss and gagestats regression types, so, in the nss call, if the gagestats reg types have come back, then call combine.

Hope that makes sense. We'll want to replicate this check (a) when the filters trigger new calls to the services and (b) in the other settings pages that combine nss and gagestats calls as well.

kjacobsen16 (changes requested):
Works well! I just made one comment with concerns about order of operations.

src/app/settings/categories/regressiontypes/regressiontypes.component.ts
```diff
@@ -99,25 +100,60 @@ export class RegressionTypesComponent implements OnInit, OnDestroy {
...
+                if (nssReturn == true && gsReturn == true) {
+                    this.combineRegressionTypes();
+                }
         });
+        this._settingsservice.getEntities(this.configSettings.gageStatsBaseURL + this.configSettings.regTypeURL+"?regions="+ this.selectedRegionID +"&statisticgroups="+ this.selectedStatisticID)
+            .subscribe(res => {
+            res.sort((a, b) => a.name.localeCompare(b.name));
+            this.gsRegressionTypes = res;
+            gsReturn = true;
+            if (nssReturn == true && gsReturn == true) {
```
kjacobsen16, line 147: Not worth changing, but you can just write `if (nssReturn && gsReturn)` since they're booleans

kjacobsen16 (approved):
Looks great!

PR #428, opened 2021-02-04
Fixes #413, Fixes #426

### LINKED ISSUE #413: Region Code Header
When adding values to calculate statistics for a region we should add the header 'Region Code' for the region code column. 

[image](https://user-images.githubusercontent.com/59063181/106656147-e698bc80-655f-11eb-8bdc-f338a1502e07.png)

From Pete's Review 2/2/2021

### LINKED ISSUE #426: Add "Regions" to statistic group label
On the NSS page after you have selected a region there are lists of regression regions but the heading is the Statistic Group associated with that regression region ex “peak-flow statistics”. Maybe the heading should instead be called “Regions with peak-flow statistics”, "Peak-Flow Statistics Regions", or something similar.

From Pete's feedback 2/2/2021

harper-wavra:
I did both issue #413 and #426 in this pull request as they were both in `mainview.component.html`.

I added "Regions" behind the statistic group names.
I added "Region Code" to the table headers.

### DIFF
--- a/src/app/mainview/mainview.component.html
+++ b/src/app/mainview/mainview.component.html
@@ -103,7 +103,7 @@
 						<th class="width-20">Explanatory Variable</th>
 						<th class="width-10 center">Code</th>
 						<th class="width-20 center">Value</th>
-						<th class="width-30" *ngIf="multipleRegRegions">Region Code</th>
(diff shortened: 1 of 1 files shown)


kjacobsen16 (approved):
👍

PR #434, opened 2021-02-17
Fixes #432

### LINKED ISSUE #432: Limit file selection in batch upload to xlsx
We should be able to set the designated file type when the user selects a file

harper-wavra:
batch-upload.component.html - added accept attribute to guide users towards selecting .xlsx or .xls files.
batch-upload.component.ts - added additional check for file type, added toast message for multiple files or incorrect file type
gs-sidebar.component.ts - removed console.log


kjacobsen16 (approved):
👍

PR #437, opened 2021-02-26
Fixes #423

### LINKED ISSUE #423: Add button to batch upload to add new row
Add button to batch upload to add new row. I quickly wrote a function and added the html code to do this. It is currently commented out. The 'new row' will be to be formatted differently based on which type of records are being submitted (i.e. stations vs stats vs chars).

harper-wavra:
src/app/gagestats/batch-upload/batch-upload.component.html - added button to add row
src/app/gagestats/batch-upload/batch-upload.component.ts  - removed unused import, added function that takes last item in `tableData` makes a "copy" of it in order to get input type then adds it to the end of `tableData`

### DIFF
--- a/src/app/gagestats/batch-upload/batch-upload.component.html
+++ b/src/app/gagestats/batch-upload/batch-upload.component.html
@@ -117,6 +117,7 @@
                     </tbody>
                 </table>
             </div>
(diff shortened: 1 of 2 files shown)


kjacobsen16 (approved):
👍

PR #438, opened 2021-02-26
Fixes #431

### LINKED ISSUE #431: Refresh main gagestats filters after submitting edits in batch upload
>When we get a response back after submitting changes, we should requery the main gagestats filters. I added some stations and it didn't update the number of pages, although it went from 730 pages to 731. Same when gages are deleted, but we can add a separate issue for that unless you find it easy to add in.

harper-wavra:
Added code to refresh stations/pages when adding bulk uploaded stations.

As for second part of issue about when gages are deleted, I added an [issue ](https://github.com/USGS-WiM/GageStatsServices/issues/98) in GageStatsServices.


kjacobsen16 (approved):
👍

PR #441, opened 2021-03-08
Fixes #439

### LINKED ISSUE #439: Table for XI row vector
The xi row vector has less info than the covariance matrix, but putting it in a table might make it a little easier on the users

harper-wavra:
Added a table for the xi row vector entry.

addscenario.component.ts
- Add new functions to add a row vector and remove a row vector
- when cloning added code to parse the cloneParameters to get number of rows and input the numbers into the table
- added code in the setUpScenario function in order to format the vector to be used in the services
- changed matrix regex to match vector regex (more robust)
- added some spacing and forgotten ;

addscenario.component.html
- Added input for rows to create a table of inputs for the vector
- added button to create and remove vector
- removed `type="number"` from matrix entry - was causing errors if user used some letters

### DIFF
--- a/src/app/shared/components/addscenario/addscenario.component.html
+++ b/src/app/shared/components/addscenario/addscenario.component.html
@@ -168,10 +168,27 @@
 								<div class="form-group">
 									<label class="req" for="xiRowVector">XI Row Vector:</label><br/>
 									<label class="example-text">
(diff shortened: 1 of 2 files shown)


kjacobsen16 (approved):
👍

PR #442, opened 2021-03-11
Fixes #435

### LINKED ISSUE #435: Gagestas batch modal updates
- [x] Stations - make "IsRegulated" required
- [x] Characteristics - make "comments" not required if possible, it's all going through as "null" I think

I have an email chain open with Ryan and Pete that might call for more items added to this issue

harper-wavra:
- For `stationChars` changed `isRegulated` to required.
- For `statChars` changed `yearsOfRecords` and `comments` to optional.
- For `charChars` changes `comments` to optional.
- Added function `checkForRequiredColumns`. Which if the headers for the batch upload have been populated it will create an array of required characteristics for each different type of upload. It will then sort that required array and the header row, compare them and find any missing required columns. If there are missing required columns it will print out an error, add to the `errorList`, and not let the user submit the data.
- Cleaned up file, spacing, missing ;, removed unused/duplicate function, removed console.logs.

### DIFF
--- a/src/app/gagestats/batch-upload/batch-upload.component.ts
+++ b/src/app/gagestats/batch-upload/batch-upload.component.ts
@@ -444,9 +444,9 @@ export class BatchUploadModal implements OnInit {
       });
       const arr1 = this.tableData[0].concat().sort();
       const arr2 = requiredChars.concat().sort();
-      let isNotMissing = (arr2.every(v => arr1.includes(v)));
-      if (!isNotMissing) {
-        this.errorList.push({'name':'missing required columns'}, {'Header': null}, {'type': null});
+      let missingColumns = (arr2.filter(x => !arr1.includes(x)));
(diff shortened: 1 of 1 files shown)


kjacobsen16 (changes requested):
Hm, I'm still able to hit the submit button when I'm (a) missing a required header (e.g. Longitude) or (b) have a gage that already exists.

harper-wavra:
@kjacobsen16  Could you provide the steps you took when submitting with missing required headers. I tried now and deleted the Longitude column and I am not able to submit the stations.
image

kjacobsen16:
This is strange.. I can't replicate the issue at all, although I can see that the station went through without a longitude 🤷‍♀️ I have no idea how I did this.

I'll call it good since I can't replicate it, but we'll have to keep an eye out and make sure we document steps if this happens again.

kjacobsen16 (approved):

PR #444, opened 2021-03-15
Fixes #422

### LINKED ISSUE #422: Update Service Endpoints for Production
When we release this for production, we'll need to point the config file to the production services (remove "test." in the services url).

We might want to think about building in something like streamstats has, where it reads the host url and determines which services to use that way: https://github.com/USGS-WiM/StreamStats/blob/dev/src/appConfig.js#L19-L28

harper-wavra:
I _think_ this will work, but I wasn't sure how to test. From my understanding when you run ng build for production it will replace `src/environments/environment.ts` with `src/environments/environment.prod.ts`, which in turn should call the newly created `assets/prodConfig.json` that has the correct service endpoints for production. I could have been way off with my thinking though, so let me know if this won't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,17 @@ Clone the repo and navigate to folder.
 
 ## Building and testing
 
+To compile the project for a production environment run:
 ```
 ng build --prod --base-href
 ```
+
+To compile the project for a non-production environment run:
+```
+ng build --base-href
+```
+
+The `--prod` tag will use the production configuration and point to the production services.
 
 ## Deployment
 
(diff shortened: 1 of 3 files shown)


kjacobsen16 (approved):
Great job! This works perfectly. I updated the read me with the different build commands for a production and non-production environment. I tested this by running each of those commands and then searching for `config.json` and checking the dist files to see which config they're using. To make sure both commands worked properly, I also uploaded the dist folder for each to a test site, but that step is unnecessary. Let me know if you'd like more information on this!

I also changed the `loginShow` and `showGageStats` to false for the production environment, since we won't need that yet.

PR #460, opened 2021-06-02
Fixes #276

### LINKED ISSUE #276: Batch Upload
An average report could have between 40-100 equations to add to NSS. If we could do something like an "Import from Excel" option, that would be helpful.

harper-wavra:
Here is my first attempt at the batch upload. It is still a little clunky, but I thought getting some initial feedback would be good. Below is a list of changes and additions to the code.

app.component.html
- Added batchUploadModalNSS
- Changed batchUploadModal to batchUploadModalGS so we can tell the different between the two.

app.module.ts
- Added BatchuploadComponentNSS
- Changed BatchuploadComponent to BatchUploadComponentGS so we can tell the different between the two.

batch-upload.component.ts & gagestats.component.ts
- renaming

mainview.component.html & mainview.component.ts 
- Add Bulk Upload button

batchupload.component.html
- Add NSS batch upload modal html
- Got some stlyes from GS batch upload
- Equation title (ex Equation 1, Equation2, ect) turns green or red if it was successfully added or not

batchupload.component.ts
- Added code to parse through the excel file
- added code to loop through equations and try to post them

batchupload.component.scss
- Added styles I got from GS batch upload

app.service.ts 
- spacing
- renaming of GS batch upload items
- Added NSS batch upload items

Added example_nss_bulk_upload.xlsx to assets folder


src/app/shared/components/batchupload/batchupload.component.html
```diff
@@ -0,0 +1,85 @@
...
+                <h6 class="title">Select a Worksheet:</h6>
+                <ng-container *ngFor="let sheetName of wb.SheetNames">
+                        <button class="button grey mright-sm mtop-xs" (click)='selectSheet(sheetName)'>{{sheetName}}</button>
+                </ng-container>
+			</div>
+
+            <div *ngIf="tableDisplay">
+                <form class="editScen">
+                    <div class="table-title" *ngFor="let e of equationData, let i = index">
+                        <h4 [ngStyle]="{'color': success[i]}">Equation {{i+1}}</h4>
```
kjacobsen16: I don't think the `[ngStyle]="{'color': success[i]}"` is working, I don't see any colors applied to these headers

src/app/shared/components/batchupload/batchupload.component.ts
```diff
@@ -0,0 +1,370 @@
...
+    this.tableDisplay = false; 
+    delete(this.data);
+  }
+
+  public selectFile(event: any) {
+    const target: DataTransfer = <DataTransfer>(event.target);
+    if (target.files.length !== 1) { //Check for multiple files
+      this.clearTable();
+      this._toasterService.pop('error', 'Error', 'Cannot select multiple files');
+      return;
```
kjacobsen16, line 161: The `multiple='false'` in the html should cover this. Since it doesn't, that indicates it's set up incorrectly. The default is to allow selection of only one file. If you remove the `multiple='false'`, it should work fine and allow only one file.

src/app/shared/components/batchupload/batchupload.component.ts
```diff
@@ -0,0 +1,370 @@
...
+    var errors: errors = {  
+      id: null,
+      value:  null 
+    };
+
+    // Loop through spreadsheet
+    for (var i = 2; i < data.length; i++) { 
+      if (data[i][0]) { // Study area
+        var studyArea = (data[i][0]);
+      } if (data[i][2]) { // Statistic Group
```
kjacobsen16: We should put these if statements on separate lines from the last one - otherwise my brain automatically thinks it's an else if/else statement.

src/app/shared/components/batchupload/batchupload.component.ts
```diff
@@ -0,0 +1,370 @@
...
+    };
+    var errors: errors = {  
+      id: null,
+      value:  null 
+    };
+
+    // Loop through spreadsheet
+    for (var i = 2; i < data.length; i++) { 
+      if (data[i][0]) { // Study area
+        var studyArea = (data[i][0]);
```
kjacobsen16: If I'm understanding correctly, I think it would make more sense to initialize these types of variables (`var studyArea`) above the loop. It seems like the loop is using whatever was last set as the studyArea, statistic group, etc. (since not every line will have them). If that's the case, we can initiate this variable above so anyone who works on this in the future is clear that this is a variable that is used in every loop, whether or not it's set in that current iteration. Hope this makes sense 😬

src/app/shared/components/batchupload/batchupload.component.ts
```diff
@@ -0,0 +1,370 @@
...
+      } if (data[i][2]) { // Statistic Group
+        var statisticGroup = (data[i][2]);
+      } if (data[i][5]) { // Regression Region
+        var regressionRegionName = (data[i][5]);
+        const regionID = this.regions.find(r => r.name == studyArea).id;
+        this.regressionRegions = await this._settingsservice.getEntities(this.configSettings.nssBaseURL + 'regressionregions?regions=' + regionID).toPromise();
```
kjacobsen16, line 236: Great use of async/await methods :) 

I do wonder if this can be optimized though. I'm thinking we can re-query this when the study area is updated instead of the regression regions. Then we're not making this call with every regression region, but with every study area, resulting in fewer service calls.

src/app/shared/components/batchupload/batchupload.component.ts
```diff
@@ -0,0 +1,370 @@
...
+      }
+      if (data[i][19]) { // Average standard error of prediction
+        errors.id = this.errors.find(et => et.code == ("ASEp")).id;
+        errors.value = data[i][19];
+        errorsArray.push(JSON.parse(JSON.stringify(errors)));
+      }
+      if (data[i][28]) { // Percent Correct
+        errors.id = this.errors.find(et => et.code == ("PC")).id;
+        errors.value = data[i][28];
+        errorsArray.push(JSON.parse(JSON.stringify(errors)));
```
kjacobsen16: Are we able to get the column name from the spreadsheet? there's a possibility of new errors being added, or the codes being changed. I think instead of getting a specific column number, we should use the column name to find the right code. This might mean that we need to change the column names to _just_ be the abbreviation, and not include the name (e.g. change "SE (Standard Error)" to just say "SE", then we can loop through the error types list from our services and look for that in the current row. We also might want to think about grabbing column names for other items, in case we add additional columns in the future

src/app/shared/components/batchupload/batchupload.component.ts
```diff
@@ -0,0 +1,370 @@
...
+    this.createTable(this.data);
+    this.tableDisplay = true;
+    this.sheetNamesButtons = false;
+  }
+
+  async createTable(data) {
+    this.success = [];
+    this.clicked = false;
+    var counter = 0;
+    var explanatoryVariablesArray: explanatoryVariables[] = [];
```
kjacobsen16: I am seeing a lot of `var` usage. Angular/Typescript prefers `const` and `let` instead. I would recommend getting ESLint set up in VSCode to catch these things.

kjacobsen16 (changes requested):
Great job! It's working for me, but, as is always the case with big updates, I do have suggestions:
- [x] This button placement is a little weird once you select a study area:
image
- [x] We should allow the user to edit their selections when we display them - this is less important than the other issues, since users can just edit the excel file and re-upload. When we do this, we'll want to make sure we add dropdowns for items that would need dropdowns.
- [ ] If there's an error, it looks like it just takes the stored bulk upload template, not my uploaded file. Because of that, I don't know how error handling and checking for good values is. In the gagestats bulk upload, Anders created a function that checks if things are valid. I think we can do that as the info is pulled into the NSS batch upload modal and identify errors and ask users to try again (or highlight the area for them to edit if we have in-modal editing set up).
- [ ] I think since we have a "Regressions" table, we can use it to display multiple equations that have the same statistic group and regression regions. This would also potentially save space, to aid in making room for #5. If instead of saying "Equation 1", "Equation 2" as the header, we name it after the regression region and/or statistics group, that's fewer columns needed in the table.
- [x] I don't see the prediction interval info in the bulk upload window after uploading my spreadsheet, but I do see it was successfully added to the DB.
- [x] What happens after a scenario is added? I think we'd want to exit the modal and reload scenarios (if a state is selected)
- [ ] Equations within the same statistic group can be merged into one object, so fewer POST calls will be needed. Pictured below are all the non-preflight calls, and there are multiple for each statistic group.
image
- [x] We eventually want users to use this to bulk _update_ scenarios as well. This is a separate task, but figured I'd mention it in case you see an easy way to allow this, or if there are better ways to organize your work to prep for it.

src/app/shared/components/batchupload/batchupload.component.html
harper-wavra: The color shows up after you click create. I added it to show the user which equations were added and which ones got an error. 
image

harper-wavra:
@kjacobsen16  For the third issue about if there is an error it taking the stored bulk upload template, do you mean when there is an error uploading a file or an error when creating the equations?

kjacobsen16:
It's when there is an error uploading equations. I tried changing a couple things in the spreadsheet, e.g. misspelling variable types in the equation or in the variables list, etc. and instead of letting me know what the error was, it just used the example template.

kjacobsen16 (approved):
I think this is good for now. Once we update this on the test server, I'll let Ryan know about the enhancements we plan to add. Hopefully this will get him what he needs for now!

@harper-wavra can you rebuild the `dist` folder before merging this? That way the dev branch has the updated dist folder. We should get in the habit of doing that. Once it's merged, we can update the nss-dev site on both the current test and new test servers. Thanks!

PR #466, opened 2021-06-30
Fixes #461

### LINKED ISSUE #461: Add Google Analytics
I thought we had this - but looks like we don't! We'll want to add the google tag info below to the index.html, and figure out what is important to document. 

We can ask Pete what he might want to see, but initial thoughts are to track when equations are computed, when scenarios are added, edited or deleted, and any other basic function. We can include the region/state and statistic groups in case Pete wants to see the frequency of those things (or maybe say "all" if no filters are applied).

```
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=G-2C1QXBBXVH"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'G-2C1QXBBXVH');
</script>
```

harper-wavra:
Added a click event to when:

- equation is computed (with region, statistic group and stat label as a label)
- citation is edited, added, or deleted
- equation is edited, added, or deleted
- regression region is edited, added, or deleted
- gage is edited, added, or deleted
- gage statistic is edited, added, or deleted
- gage characteristic is edited, added, or deleted

### DIFF
--- a/src/app/gagestats/add-station/add-station.component.ts
+++ b/src/app/gagestats/add-station/add-station.component.ts
@@ -10,6 +10,7 @@ import { Agency } from 'app/shared/interfaces/agency';
 import { Region } from  'app/shared/interfaces/region';
 import { StationType } from 'app/shared/interfaces/stationtypes';
 import { HttpParams } from '@angular/common/http';
+declare let gtag: Function;
 
 @Component({
   selector: 'addStationModal',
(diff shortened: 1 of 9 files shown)


kjacobsen16 (approved):
Looks good! I made some small changes and added an issue to add more details to the notifications for the management side. Since that's not going to be published yet, I marked it as an enhancement.

Issue #496, opened 2021-07-13
Some weird printing with the equations


kjacobsen16:
To replicate:
- In the NSS page, select "Alaska"
- Put example values in for all of the parameters (in order, my values were 1500, 50, 3600, 20, 20, 20)
- Hit Compute
- Scroll to bottom and select "print report"
- Scroll down to the Appendix section - it takes a minute to load all the equations in mathjax format, but this issue displays whether or not the equations are formatted yet
![image](https://user-images.githubusercontent.com/35497731/125530825-06e8a8a1-6e15-492f-9c4f-18a5a5b65c7b.png)
![image](https://user-images.githubusercontent.com/35497731/125531114-f74409ee-f670-4cdf-a539-0ce58cbca50d.png)

PR #507, opened 2021-08-25
Fixed appendix table header printing issue #496

mitchas:
I added some CSS to break up the print display a bit to prevent the header printing issue #496 

I tested it using the values in that issue, as well as a few others and it seems to be fixed. 

If someone could try a few more and double-check, then it's probably good to merge.


harper-wavra (approved):
Looks good, thanks!

PR #515, opened 2021-11-08
unprotect sheets

harper-wavra:
Unprotect the bulkupload gagestats excel sheet in order to update the data in the sheet.


amedenblik (changes requested):
~~I opened the updated Excel file, clicked "Data" > "Refresh all", and got this error. It seems that unprotecting the sheets did not carry over. I wonder if this is due to changes in Excel from the latest update.~~ File is fine, sorry!
image

Updated Excel file also needs to be in the src folder.

harper-wavra:
I think it wasnt working before because it was pulling the excel sheet from src/assets which wasnt updated, let me know if this works.

amedenblik (approved):

PR #516, opened 2021-11-29
Fixes #514

### LINKED ISSUE #514: Add citation option with bulk upload statistics
There is no way currently to add statistics with citation linked in bulk upload this is necessary in order to add multiple of the same statistic types to a gage, assuming the unit types and statistics wouldn’t change. Currently you can add more than one of the same statistic type to a gage as long as citation, unit type, or statistic group are different.

harper-wavra:
You are now able to add statistics and characteristics with associated with them citation linked when using the bulk upload feature. 
- Edited the example_gagestats_bulk_upload template. I had to add citations to the data sheets and create a spot for citations to be added to the statistic and characteristic sheets.
- There was a way to add a single citation to all the statistics or characteristics you were uploading. I removed this as this wont be needed.
- Fixed an issue when nothing was being set through it was using "null" instead of null which caused errors.
- Fixed an issue with the loader not turning off if there was an error with the upload.
- Rebuilt the application.


amedenblik (changes requested):
New spreadsheet does not appear to successfully allow multiple Characteristics to be uploaded for a single station, even if the citations are different.

image

image

amedenblik (approved):
There is a remaining issue where you cannot upload multiple Characteristics for a single station if they have the same variable type and units. This has been moved to [GageStatServices Issue #112](https://github.com/USGS-WiM/GageStatsServices/issues/112).

PR #528, opened 2022-03-08
Updates to bulk upload

harper-wavra:
- Reworked how its getting `stationIDs`. 
    - Before you were able to submit the stats and chars before all the `stationIDs` were returned, resulting in an error. I added `async and await` to wait for all the `stationIDs` to be returned before you can select submit.  
    - Before it was grabbing every `stationID` for every station even if it was a repeat, this was taking a lot of time. I reworked it to only make a request to get a `stationID` if that station had not been used yet. 
- The toast message was confusing when there was an error. The services response with an info message saying it uploaded so many stations but that is false. I edited the message to only show the error and not the info message.
- Rebuilt the dist folders

I was able to test with actual data from Ryan. I was able to upload 200 chars and stats at a time with no issues.


amedenblik:
Sample upload

[example_gagestats_bulk_upload.xlsx](https://github.com/USGS-WiM/NSS/files/8234103/example_gagestats_bulk_upload.xlsx)

amedenblik (approved):

PR #559, opened 2022-11-02
Fixes #506

### LINKED ISSUE #506: Assign unit types and statistic group types for the regression types
We want to also allow users in NSS to assign the metric unit type, english unit type and statistic group type for the regression types when adding or editing.

There are also some other places where we might want to pull this info, such as the add scenario component - that can be used to more easily (a) filter the regression types by statistic group (in the past the filters were based on if the regression type was used in an equation assigned to a statistic group, not based on the default statistic group for that regression type), and (b) to get the unit options for the scenario they're adding. It also might be helpful to check the default units and statistic groups for regression types on equations entered in the batch upload in NSS. Note that this might cause some confusion with the filters on the NSS endpoint - since those filters are based on the usage in equations, not the assigned default values.

harper-wavra:
Allows users to edit or add unit types and statistics groups to regression types. 
Displays unit type and statistic group for regression types.

Ryan was doing some work in NSS and we noticed this should be complete before he adds more regression types.


src/app/settings/categories/regressiontypes/regressiontypes.component.html
```diff
@@ -29,9 +29,12 @@ <h3 class="mbottom-md">Regression Types</h3>
...
                 <tr>
                     <th class="width-5"></th>
-                    <th class="width-40">Name</th>
-                    <th class="width-40">Description</th>
+                    <th class="width-25">Name</th>
+                    <th class="width-25">Description</th>
                     <th class="width-15">Code</th>
+                    <th class="width-10">Metric Unit</th>
+                    <th class="width-10">English Unit</th>
+                    <th class="width-10">Statstic Group</th>
```
amedenblik: Typo: `statstic`

CHANGELOG.md
```diff
@@ -13,12 +13,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added "Limit by regulation status" filter in sidebar
 - Added equivalent years of record, drainage area export, and order index to equations table 
 - Allow user to sort Variable Types by Statistic Group in Settings
+- Allow user to add unit types and statstic groups to regression types
```
amedenblik: Typo: `statstic`

amedenblik (changes requested):
Works well!

amedenblik (approved):

